PR #1, opened 2017-08-21
Fix deprecated BindTo usage in ttyrun-getty@.service.in.

pkern:
BindTo was renamed to BindsTo in systemd back in v187 from July 2012.

Signed-off-by: Philipp Kern <pkern@debian.org>

### DIFF
--- a/systemd/ttyrun-getty@.service.in
+++ b/systemd/ttyrun-getty@.service.in
@@ -11,7 +11,7 @@
 [Unit]
 Description=TTYRun on %I
 Documentation=man:ttyrun(8) man:agetty(8)
-BindTo=dev-%i.device
+BindsTo=dev-%i.device
 After=dev-%i.device systemd-user-sessions.service plymouth-quit-wait.service
 After=rc-local.service
 Before=getty.target


pkern:
FWIW, this was based on [Debian bug #857190](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=857190) which pointed to https://github.com/systemd/systemd/commit/7f2cddae09fd2579ae24434df577bb5e5a157d86 as the commit that changed this setting in systemd.

michael-holzheu:
@hbrueckner : Could you please check?

michael-holzheu:
@pkern : If we change "BindTo" to "BindsTo" we would loose backward compatibility for all systems hat have systemd < v187, right?

Are there any plans to remove "BindTo" from systemd? In case it is only an aesthetic issue, I am not sure if we should change the name now.

hbrueckner:
The rename to BindsTo goes really back to 2012 with systemd commit:

commit 7f2cddae09fd2579ae24434df577bb5e5a157d86
Author: Lennart Poettering <lennart@poettering.net>
Date:   Fri Jul 13 23:34:40 2012 +0200

    unit: rename BindTo= to BindsTo=
    
    all other dependencies are in 3rd person. Change BindTo= accordingly to
    BindsTo=.
    
    Of course, the dependency is widely used, hence we parse the old name
    too for compatibility.

Given a grace period of 5 years, the proposed change is fine with me.

@Phil: Many thanks for your contribution!

michael-holzheu:
@pkern : Applied - Congratulations for the first external contribution to s390-tools :1st_place_medal:

PR #3, opened 2017-08-25
zipl: Always build and link without PIE.

pkern:
Compilers in Debian and Ubuntu default to enabling PIE for all binaries.
This is a poor fit for boot loader code that has strict requirements
for its layout. Hence always forcefully disable PIE.

Signed-off-by: Philipp Kern <pkern@debian.org>

### DIFF
--- a/common.mak
+++ b/common.mak
@@ -218,6 +218,16 @@ endif
 export AS LD CC CPP AR NM STRIP OBJCOPY OBJDUMP INSTALL CFLAGS CXXFLAGS \
        LDFLAGS CPPFLAGS ALL_CFLAGS ALL_CXXFLAGS ALL_LDFLAGS ALL_CPPFLAGS
 
+ifneq ($(shell $(CC_SILENT) -dumpspecs 2>/dev/null | grep -e '[^f]no-pie'),)
+	NO_PIE_CFLAGS := -fno-pie
+	NO_PIE_LINKFLAGS := -no-pie
+	NO_PIE_LDFLAGS := -no-pie
+else
+	NO_PIE_CFLAGS :=
+	NO_PIE_LINKFLAGS :=
+	NO_PIE_LDFLAGS :=
+endif
+
 # Overwrite implicite makefile rules for having nice compile output
 %.o: %.c
 	$(CC) $(ALL_CPPFLAGS) $(ALL_CFLAGS) -c $< -o $@
--- a/zipl/boot/Makefile
+++ b/zipl/boot/Makefile
@@ -1,7 +1,7 @@
 # Common definitions
 include ../../common.mak
 
-CFLAGS_BOOT = -Os -g -I../include -D__ASSEMBLY__ \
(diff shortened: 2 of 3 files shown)


michael-holzheu:
@oberpar : Could you have a look, please.

michael-holzheu:
Assigning to @stefan-haberland (back from vacation)

stefan-haberland:
I had pretty much  the same patch in our tree but unfortunately this does not build with older compiler versions which are not aware of the no-pie options. So I removed it and wanted to add some kind of check for the availability of the -pie or -no-pie option to allow the s390-tools to be build with older compiler versions. But unfortunately I did not have the time to find a proper solution yet.

Do you have an idea for this?

pkern:
Do you know which old compilers broke? (I don't dispute that, I just want to know the baseline.) Of course it's harder to just drop-in autodetection as built by others as s390-tools does not use any pre-existing build system.

xnox:
I believe gcc-5 has the -no-pie flag, and prior to that did not. I can double check with doko.

pkern:
Please check the followup patch I just added. That uses gcc's -dumpspecs output. I don't know if that's more acceptable. Please test it with whatever ancient compiler you still need to support. In my test with a (non-cross) gcc-4.9 it correctly omitted no-pie while it put it there with a somewhat current gcc.

stefan-haberland:
Thanks for the patch. In theory this looks perfectly fine but for whatever reason it does not work on a SLES12SP2 for example (gcc 4.8.5). If I build with "V=1" it works but leaving it at its default gcc fails with:

LINK    zipl/boot/fba0.exec
gcc: error: unrecognized command line option ‘-no-pie’
Makefile:50: recipe for target 'fba0.exec' failed
make[2]: *** [fba0.exec] Error 1

I need to have a look at this why exactly the non verbose compile does not work.

stefan-haberland:
The reason is that the

$(shell $(CC) -dumpspecs 2>/dev/null | grep -e '[^f]no-pie')

gets translated into

("/bin/sh", ["/bin/sh", "-c", "echo \"  CC      \"zipl/;gcc -dumpspecs 2>/dev/null | grep -e '[^f]no-pie'"], [/* 80 vars */]) = 0

I think with the "echo" in it the ifneq statement  is always true and therefore -fno-pie is always added.

With the V=1 option it is translated into:

("/bin/sh", ["/bin/sh", "-c", "gcc -dumpspecs 2>/dev/null | grep -e '[^f]no-pie'"], [/* 80 vars */]) = 0

and this works as expected.

I suggest to use $(CC_SILENT) instead of $(CC). I have tested this on SLES12 and Ubuntu 16.04/16.10..

pkern:
Done. Thanks!

stefan-haberland:
Can you please merge the three commits into one? Afterwards I will apply it to the tree. Thanks!

pkern:
Done.

PR #8, opened 2017-10-12
dasdfmt cleanup v2

hreinecke:
Hi all,

here's the second version of my dasdfmt cleanup, should address all issues raised with v1.
Please apply.

Cheers,

Hannes


hoeppnerj:
Patches look good and have been merged internally. They'll be pushed to master with the next sync.

Thanks for the contribution!

so long,
Jan

PR #9, opened 2017-10-12
Update gitignore to not exclude init.d and rules.d directories

hreinecke:
Signed-off-by: Hannes Reinecke <hare@suse.com>


michael-holzheu:
Applied - thanks for the separate PR.

PR #10, opened 2017-10-12
zipl: remove invalid dasdview command line option

r4f4:
dasdview command does not accept the '-f' option so because of that the
command will always fail.

### DIFF
--- a/zipl/src/zipl_helper.device-mapper
+++ b/zipl/src/zipl_helper.device-mapper
@@ -623,7 +623,7 @@ sub get_dasd_info($)
 	my $type;
 	local *HANDLE;
 
-	open(HANDLE, "$dasdview -x -f  $dev 2>/dev/null|") or
+	open(HANDLE, "$dasdview -x $dev 2>/dev/null|") or
 		# dasdview returned with an error
 		return undef;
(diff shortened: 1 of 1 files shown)


xnox:
We have discovered this in ubuntu now. This means that since the removal of the flag in v1.38 or v1.39? systems were not able to boot when installed on top of multipath devices, or on top of LVM on DASD drives. This kind of indicates that s390-tools releases go out untested?

michael-holzheu:
> dasdview command does not accept the '-f' option so because of that the command will always fail.

The -f option for dasdview and dasdfmt has been removed in [s390-tools-1.37.1](https://www.ibm.com/developerworks/linux/linux390/s390-tools-1.37.1.html). Looks like the zipl-helper script has been forgotten - I apologize for that.

@hoeppnerj @oberpar @stefan-haberland : Can you please have a look?

michael-holzheu:
> This kind of indicates that s390-tools releases go out untested?

We do several internal tests but obviously we did not test the LVM zipl scenario.

stefan-haberland:
Thanks for the patch. Looks good to me.

Acked-by: Stefan Haberland <sth@linux.vnet.ibm.com>

Issue #5, opened 2017-09-04
RFE: get rid of Perl based tools to reduce dependencies


sharkcz:
While working on the "Modular Fedora" project, where one of the goals to reduce the size of the installed content, the team realized that there are some tools written in Perl dragging it into the system. I think the most prominent one is the zipl_helper.device-mapper helper script in zipl, but there are few more.

michael-holzheu:
@sharkcz : Could you please provide some more information / links on "Modular Fedora"?
Aren't there thousands of perl programs for Linux. Do you want to eliminate them all?

michael-holzheu:
@oberpar : FYI - since zipl helper explicitly has been mentioned here.

sharkcz:
The idea is to start with a rather minimal system and then add content on top of that, you can find more at https://github.com/fedora-modularity/hp and https://fedoraproject.org/wiki/Modularity (and more). As the bootloader is always part of the minimal system (on all arches) to make system usable at all it's now dragging Perl into the package set (on s390x). Similar problem on ppc is tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1463749 There is no problem with Perl in "upper" layers.

michael-holzheu:
Ok, so your goal is that the [base-runtime](https://github.com/fedora-modularity/base-runtime) (?) minimal system does not contain Perl.

Currently for RHEL7.4 we have the following Perl programs in s390utils-base:
```
$ grep -r "usr/bin/perl" *
lib/s390-tools/zipl_helper.device-mapper:#!/usr/bin/perl -w
lib/s390-tools/cpumf_helper:#!/usr/bin/perl -W
sbin/zfcpdbf:#!/usr/bin/perl
usr/sbin/ip_watcher.pl:#!/usr/bin/perl -w
usr/sbin/chmem:#!/usr/bin/perl
usr/sbin/lsmem:#!/usr/bin/perl
usr/sbin/chcpumf:#!/usr/bin/perl -W
usr/sbin/lsluns:#!/usr/bin/perl
usr/bin/lscpumf:#!/usr/bin/perl -W
```

What should be part of the base-runtime? Only glibc?

sharkcz:
> Ok, so your goal is that the base-runtime (?) minimal system does not contain Perl.

right, respectively removing Perl, which is not used otherwise, is a way to reach the goal

> What should be part of the base-runtime? Only glibc?

[...]

PR #12, opened 2017-10-17
WIP: #5: rewrite zipl_helper in python.

r4f4:
This is still a work in progress regarding issue #5. The C version would take longer than expected, so I did a python one in the meantime. Using python is still in line with the Fedora minimal baseruntime, since it is a dependency of some basic system tools (like dnf).

I tried to included some unit tests to make sure the helper works as expected. I'll continue testing and improving the test cases while upstream reviews this patch.

I would love some feedback before submitting the definitive version.

### DIFF
--- a/zipl/src/zipl_helper.device-mapper
+++ b/zipl/src/zipl_helper.device-mapper
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env python3
 #
 # zipl_helper.device-mapper: print zipl parameters for a device-mapper device
 #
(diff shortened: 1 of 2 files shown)


r4f4:
ping. Any comments on this?

michael-holzheu:
> ping. Any comments on this?

@r4f4 - sorry for the late reply.

Our strategy for s390-tools is to use C as the main language. Therefore we started converting tools (lsqeth, lsscm, lschp, chchp) to C and started with the internal "libutil" support library. Having only one language makes support much easier since all of our developers know C and sharing code is possible.

Up to now we don't have python in s390-tools. So if we go your way, this would add yet another language to the package which I really would like to avoid.

Therefore I would appreciate it if you could try to re-implement the script in C.

michael-holzheu:
@oberpar @stefan-haberland @hoeppnerj : FYI

r4f4:
Ok. I'll keep working on the C version.

michael-holzheu:
> Ok. I'll keep working on the C version.

Great :+1:

PR #13, opened 2017-10-20
vmur changes

donald-russell:
Hello,
I have have made some changes to the vmur program for working with zVM ur devices, specifically print and punch files. 

I added a --wait option so vmur will wiat for the lock file to become free, rather than show a "vmur in use" error.
I added a --tag option to allow more control over the tag data, it cannot be used with --node option.
I changed the lock file name to include the device number so different instances of vmur using different devices can proceed.
I added code to prevent command-injection and unwittingly running an unintended CP command. 
I added info to the man file for the options I added.

I'd appreciate your feedback on my changes and would like you include them in future releases of the s390-tools package.

The reason I made these changes is because I have linux processes that create JCL for submission to zOS systems. Though I could get the JCL there using --node MVSX --user JOB, that's not really accurate, and there's no mechanism to supply priority or RSCS options. Using the --tag option I can punch a file to --user RSCS --tag 'mvsx job 10 ( accmsg=no enqmsg=no finalmsg=no sentmsg=no )'

and of course the --wait option allows multiple processes submitting JCL to both run simulataneously and serialize properly on the using the virtual punch device.

Normally I submit the JCL to zOS over FTP to JES, but RSCS is a backup mechanism in case the FTP fails for some reason. It's all automated, so I can't have one submission work and another error out with "vmur is in use" :-)

Thanks for your consideration.
Donald Russell
Visa Inc

### DIFF
--- a/vmur/vmur.8
+++ b/vmur/vmur.8
@@ -201,7 +201,7 @@ Synopsis:
 .IP "" 2
 punch|print [-fr] [-d dev_node]
 [-C class] [--form form] [--dest dest] [--dist distcode]
-[-t | -b sep.pad] [-u user] [-n node]
+[-t | -b sep.pad] [-u user] [-n node] [--tag data] [--wait]
 .br
       [-N name.type] [file]
 .PP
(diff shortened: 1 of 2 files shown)


michael-holzheu:
Duplicate of: PR #14

PR #16, opened 2017-11-13
Improve basics in netboot example

cpaelzer:
Hi,
I know the mk-pxelinux-ramfs is menat as an example and note e.g. to implement full syslinux config.
But while testing with it there were a few thing that I considered low hanging fruits and therefore wanted to provide back to you.

It is mostly about the transparency of the PXE process for a user starting from it as well as fixing up some issues around the PXE config handling.

### DIFF
--- a/netboot/mk-pxelinux-ramfs
+++ b/netboot/mk-pxelinux-ramfs
@@ -164,7 +164,6 @@ echo "$cmd: Making init script"
 # --- begin init script
 cat <<'EOF' > $builddir/init
 #!/bin/sh
-script="/sbin/pxeboot.script"
 /bin/mount -t devtmpfs none /dev
 /bin/mount -t proc none /proc
 /bin/mount -t sysfs none /sys
@@ -173,12 +172,7 @@ script="/sbin/pxeboot.script"
 /sbin/modprobe virtio_net
 /sbin/udhcpc -O pxeconffile -O pxepathprefix &
 
-while true
-do
-	${script}
-	/bin/sleep 1
-	echo "retry ${script}"
-done
+/sbin/pxeboot.script
 EOF
 # --- end init script
 chmod +x $builddir/init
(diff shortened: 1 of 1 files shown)


netboot/mk-pxelinux-ramfs
```diff
@@ -174,8 +175,9 @@ cat <<'EOF' > $builddir/init
 
 while true
 do
-	/sbin/pxeboot.script
+	${script}
 	/bin/sleep 1
+	echo "retry ${script}"
```
mihajlov: In hindsight, looping for ever seems like a bad idea ... the only reason to retry is because we wait for the DHCP response. We might consider to wait for the DHCP from within the script.

netboot/mk-pxelinux-ramfs
```diff
@@ -347,6 +349,7 @@ then
     set +e
     for c in $CONFIGS
     do
+	echo "fetching config pxelinux.cfg/$c from $siaddr"
```
mihajlov: In my case tftp was echoing all the attempted file names. This isn't the case for you?

netboot/mk-pxelinux-ramfs
cpaelzer: I'm ok with not looping (or waiting for dhcp) or a limited amount of retries, do you want me to submit such a change as part of the series?

netboot/mk-pxelinux-ramfs
cpaelzer: No,
it does not echo anything, without the fixes it only looks like:

tftp: timeout
tftp: timeout
tftp: timeout
tftp: timeout

Knowing what and where it tries to fetch helps a lot to understand what is going on.

netboot/mk-pxelinux-ramfs
cpaelzer: Per IRC discussion I pushed a change to this PR that makes it a timeout based check.

cpaelzer:
Rebasing as discussed on IRC to add Author correctly and to merge the two verbosity changes.
Pushing now...

michael-holzheu:
Applied, thanks!

PR #21, opened 2017-11-24
ziomon: fix typo

r4f4:

### DIFF
--- a/ziomon/ziomon_util.c
+++ b/ziomon/ziomon_util.c
@@ -662,7 +662,7 @@ static int init_adapters(struct adapters *all_adapters, struct options *opts)
 				malloc(sizeof(char)*MAX_HOST_PATH_LEN);
 			sprintf(adapter->q_full_path, "%s/queue_full",
 				opts->host_path[i]);
-			if (stat(adapter->path, &buf)) {
+			if (stat(adapter->q_full_path, &buf)) {
 				fprintf(stderr, "%s: Path does not exist: %s - correct kernel version?\n",
 					toolname, adapter->q_full_path);
 				rc++;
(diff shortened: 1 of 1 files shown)


michael-holzheu:
@r4f4 Thanks for the fix, but from the patch (at least) I don't understand what exactly the problem/symptom is.

r4f4:
In the case of `check_host_param' and `init_adapters', it won't detect the absence of 'queue_full' in an old kernel installation since the stat call will always succeed at that point because it's checking against a path that was already verified to exist. Now, if that's going to cause any visible symptoms I can't tell since I'm not familiar enough with the code.

Benjamin-Block:
Nice catch, these seem to be copy&paste mistakes. I think the changes are fine, but Steffen is the maintainer of the tools.

There is no impact on how the tools work, but in case the 'queue_full' attribute doesn't exist for a very old kernels, they make ziomon exit gracefully.

r4f4:
I'll update the patch with a rationale if you think it's worth merging.

Benjamin-Block:
Well, as long as the code is still in ziomon - and I don't see it disappearing anytime soon - it would be good to have it correct :-). Even though I'm not sure anyone still uses such old kernels where this attribute doesn't already exist.

r4f4:
I updated the patch description. Thank you for looking into it @michael-holzheu and @Benjamin-Block.

michael-holzheu:
> I updated the patch description. Thank you for looking into it @michael-holzheu and @Benjamin-Block.

Looks good to me now :thumbsup:, thanks!

steffen-maier:
@r4f4 thanks for the fix, will integrate it

PR #22, opened 2017-12-07
zconf: rewrite lsluns in C

r4f4:
This is one step further regarding issue #5


michael-holzheu:
@Benjamin-Block Could you please have a look?

r4f4:
Any comments on this patch?

Benjamin-Block:
Hej Rafael. Sry, but our team (zFCP) doesn't have enough room to review full tool rewrites. We added this to our backlog, but it may take a while till someone can work on it; I can't give you a time frame for this ATM.

I don't feel comfortable with too much code-churn at once. As is, these tools take some effort servicing for us. Reviews and verification on this 'scale' will take effort we can't spend right now.

Regarding issue #5; if fedora wants to get rid of perl dependencies that are pulled in because of core-packages, we have no issue if they decide to remove lsluns from s390utils-base, and put it into a subpackage. Much like they already do with various other tools like ziomon, cpuplugd, ... Then our users could install them still, and would only then pull perl (which I assume is fine, because I doubt fedora wants to get rid of perl in *every* package). As far as I am aware no other tools depend directly on lsluns.

r4f4:
@Benjamin-Block I understand. Thank you for the reply. The most important rewrite is the zipl_helper, since it's needed for booting.The others can take some time or not happen at all in case there is no interest.

PR #23, opened 2017-12-11
ziomon: rewrite fcpconf in bash script.

r4f4:
This is one step further regarding issue #5

### DIFF
--- a/ziomon/ziomon_fcpconf
+++ b/ziomon/ziomon_fcpconf
@@ -1,38 +1,29 @@
-#!/usr/bin/perl
+#!/bin/bash
 #
 # FCP adapter trace utility
 #
 # Script to collect all system information required for analysis
 #
 # Copyright IBM Corp. 2008, 2017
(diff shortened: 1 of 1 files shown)


stefan-haberland:
@Benjamin-Block @steffen-maier Could you please have a look?

steffen-maier:
ziomon is already in a subpackage and ziomon is not required for the base system, so we don't seem to require this for resolving issue #5.
We are aware that ziomon has a list of requirements [rpm -q --changelog s390utils-ziomon | grep -e '^\*' -e zio]:

* Wed Jul 23 2014 ... - 2:1.23.0-13
- update Requires for s390utils-ziomon subpackage (#1079474)

and especially

* Tue Jun 17 2014 ... - 2:1.8.2-54
- update Requires for s390utils-ziomon subpackage (#1109898)

to fix a real user issue.
I'd rather not unnecessarily re-write core performance collection parts of ziomon, because of regression risk.

PR #31, opened 2018-06-19
Adjust zdev modprobe path to be compatible with split-usr systems.

xnox:
mk-pxelinux-ramfs, mod_fsstatd|procd.service, zfcpdump all use
/sbin/modprobe path. Adjust zdev to use that path as well. This works
universally on both usr-merge systems (e.g. Debian/Ubuntu with
usrmerge package installed), and split-usr systems alike
(Debian/Ubuntu current stable releases default).

Bug-Launchpad: https://bugs.launchpad.net/ubuntu/+source/s390-tools/+bug/1777600

### DIFF
--- a/zdev/include/path.h
+++ b/zdev/include/path.h
@@ -20,7 +20,7 @@
 #define	UDEV_SUFFIX		".rules"
 
 #define	PATH_MODPROBE_CONF	"/etc/modprobe.d"
-#define PATH_MODPROBE		"/usr/sbin/modprobe"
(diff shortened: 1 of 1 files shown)


oberpar:
Thanks for the fix proposal.

I also checked the main current non-Debian based distributions and /sbin/modprobe seems to be universally available. So this is a working approach to fix the problem.

@hoeppnerj: Can you integrated this fix?

hoeppnerj:
> @hoeppnerj: Can you integrated this fix?

Will do!

@xnox Thanks for the contribution!

PR #38, opened 2018-10-02
cpacfstatsd.service: only run on LPARs.

xnox:
Add ConditionVirtualization=no to the service file such that it is
only started on LPARs. This service cannot run on z/VM nor KVM, and
fails resulting in a degraded boot.

### DIFF
--- a/systemd/cpacfstatsd.service.in
+++ b/systemd/cpacfstatsd.service.in
@@ -10,6 +10,7 @@
 [Unit]
 Description=CPACF statistics collection daemon process for Linux on System z
 Documentation=man:cpacfstatsd(8)
+ConditionVirtualization=no
 
 [Service]
 Type=forking


hoeppnerj:
Thanks for the contribution. Already pulled and integrated internally.

xnox:
More info on ConditionVirtualization from https://www.freedesktop.org/software/systemd/man/systemd.unit.html


> ConditionVirtualization= may be used to check whether the system is executed in a virtualized environment and optionally test whether it is a specific implementation. Takes either boolean value to check if being executed in any virtualized environment, or one of vm and container to test against a generic type of virtualization solution, or one of qemu, kvm, zvm, vmware, microsoft, oracle, xen, bochs, uml, bhyve, qnx, openvz, lxc, lxc-libvirt, systemd-nspawn, docker, rkt to test against a specific implementation, or private-users to check whether we are running in a user namespace. See systemd-detect-virt(1) for a full list of known virtualization technologies and their identifiers. If multiple virtualization technologies are nested, only the innermost is considered. The test may be negated by prepending an exclamation mark.

PR #41, opened 2018-10-02
zdev/initramfs: add s390-tools- prefix to hook, due to conflicts.

xnox:
Unfortunately zdev hook already exists in Ubuntu, from an unrelated
project. ZFS uses zdev/zpool names, and ships a zdev hook to do ZFS
specific initialisation. It is available on s390x and thus results in
file-conflict upon installing both. Thus renaming this zdev hook to
s390-tools-zdev.


hoeppnerj:
This would be fine with me but I'd like to have @oberpar opinion as well before I pull.

hoeppnerj:
Thanks for the contribution. Pulled and integrated internally.

PR #42, opened 2018-10-02
zkey: add initramfs hook.

xnox:

### DIFF
--- a/zkey/initramfs/Makefile
+++ b/zkey/initramfs/Makefile
@@ -1,0 +1,21 @@
+# Common definitions
+include ../../common.mak
+
+INITRAMFSDIR := /usr/share/initramfs-tools
+HOOKDIR := $(INITRAMFSDIR)/hooks
+INITTOP := $(INITRAMFSDIR)/scripts/init-top
+
+# HAVE_INITRAMFS
+#
+# This install time parameter determines whether the zkey initramfs support is
+# installed (HAVE_INITRAMFS=1) or not (default). When installed, the module
+# performs the following functions when mkinitramfs is run:
+#
+#  - install a hook to include zkey related utilities and zkey repository in
+#    the initramfs
+#
+ifeq ($(HAVE_INITRAMFS),1)
+install:
+	$(INSTALL) -m 755 -d $(DESTDIR)/$(HOOKDIR) $(DESTDIR)/$(INITTOP)
+	$(INSTALL) -m 755 hooks/s390-tools-zkey $(DESTDIR)/$(HOOKDIR)
+endif
--- a/zkey/initramfs/hooks/s390-tools-zkey
+++ b/zkey/initramfs/hooks/s390-tools-zkey
@@ -1,0 +1,45 @@
+#!/bin/sh
+#
+# Copyright IBM Corp. 2016, 2017
+# Copyright Canonical Ltd 2018
+#
+# s390-tools is free software; you can redistribute it and/or modify
+# it under the terms of the MIT license. See LICENSE for details.
+#
+# hooks/s390-tools-zkey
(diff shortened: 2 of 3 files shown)


ifranzki:
Can you tell if this change is distribution independent? 
We probably would not want to include something that works only for a certain distribution.

hoeppnerj:
@xnox can you provide some input here? Thanks!

frank-heimes:
Just caught up with @xnox and we think that this is distribution agnostic and should be pretty straight forward.
From our point of view this can and should go upstream.

zkey/initramfs/Makefile
```diff
@@ -0,0 +1,21 @@
...
+# This install time parameter determines whether the zkey initramfs support is
+# installed (HAVE_INITRAMFS=1) or not (default). When installed, the module
+# performs the following functions when mkinitramfs is run:
+#
+#  - install a hook to include zkey related utilities and zkey repository in
+#    the initramfs
+#
+ifeq ($(HAVE_INITRAMFS),1)
+install:
+	$(INSTALL) -m 755 -d $(DESTDIR)/$(HOOKDIR) $(DESTDIR)/$(INITTOP)
```
mhartmay, line 19:
```suggestion
	$(INSTALL) -m 755 -d $(DESTDIR)$(HOOKDIR) $(DESTDIR)/$(INITTOP)
```
No `/` after `$(DESTDIR)` to allow relative install paths. Same applies to the other occurrences of `$(DESTDIR)`.

zkey/initramfs/hooks/s390-tools-zkey
```diff
@@ -0,0 +1,45 @@
...
+
+for x in $zdev_modules ; do
+	manual_add_modules ${x}
+done
+
+# copy utils
+copy_exec /sbin/chzcrypt
+copy_exec /sbin/lszcrypt
+copy_exec /usr/bin/zkey
+copy_exec /usr/bin/zkey-cryptsetup
```
ifranzki, line 42: zkey-cryptsetup has a dependency on libcryptsetup (and possibly other libraries, i.e. openssl). How is this handled in regards to initramfs ?
We might also want to copy the zkey plugins (EKMFWeb and KMIP), but those have further dependencies, too.

frank-heimes: libcryptsetup got added to the initramfs
since it's with ~ 480k not that big, for other we didn't saw a direct need (aot).

zkey/initramfs/Makefile
frank-heimes: Looks reasonable to allow relative install paths.
(I just don't have permissions to update this PR)

hoeppnerj:
@frank-heimes How likely is it that @xnox will be updating the PR? If he's too busy with other things we can pull as is and adapt our requested changes ourselves. I don't think it makes sense to wait any longer if everyone agrees on the general inclusion of this PR.

frank-heimes:
I think it's best to pull as is and adapt further changes on top - imho.

zkey/initramfs/hooks/s390-tools-zkey
xnox: copy_exec copies binary, and any/all dependencies said binary needs. Just like in any other initrd implementations out there.

It doesn't copy dlopen() things however, so if we desire plugins we could add them too. But are those required to unlock drives? the intention of the initramfs hook to provide useful binaries for recovery and unlocking partitions. No need to support setting up brand new encrypted partitions for example.

xnox:
> Can you tell if this change is distribution independent? We probably would not want to include something that works only for a certain distribution.

These changes are suitable for distributes that use initramfs-tools & ship zkey.

PR #44, opened 2018-10-16
zdev: Trigger generic_ccw devices on any kernel module loads.

xnox:
Generic CCW device can use any driver, and the value of the driver is
not known ahead of time. To avoid the race between loading and binding
a kernel module, and devices added - retrigger generic-ccw devices on
any kernel module load.

Fixes https://github.com/ibm-s390-tools/s390-tools/issues/37


xnox:
nah, this is broken.

PR #45, opened 2018-10-16
zdev: Trigger generic_ccw devices on any kernel module loads.

xnox:
Generic CCW device can use any driver, and the value of the driver is
not known ahead of time. To avoid the race between loading and binding
a kernel module, and devices added - retrigger generic-ccw devices on
any kernel module load.

Fixes https://github.com/ibm-s390-tools/s390-tools/issues/37
LP: #1794308

### DIFF
--- a/zdev/include/ccw.h
+++ b/zdev/include/ccw.h
@@ -62,10 +62,12 @@ struct ccw_devinfo {
  * ccw_subtype_data - CCW subtype specific information
  * @ccwdrv: The name of the CCW device driver for this subtype
  * @mod: The name of the main kernel module for this subtype
+ * @any_driver: If set, the exact driver for this subtype are not known
  */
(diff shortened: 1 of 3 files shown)


zdev/include/ccw.h
```diff
@@ -66,6 +66,7 @@ struct ccw_devinfo {
 struct ccw_subtype_data {
 	const char *ccwdrv;
 	const char *mod;
+	const bool any_driver;
```
oberpar: The `const` here is not necessary and will likely only confuse readers, so please remove it. Also please add a description of the new field to the comment preceding the struct. Something like

`@any_driver: If set, the exact driver for this subtype are not known. `

hoeppnerj:
Thanks for the contribution. Pulled and integrated internally.

PR #50, opened 2018-12-21
Automatically bring up devices in the initrd for Z13.

ltrager:


ltrager:
This patch is needed to add support for booting the Ubuntu ephemeral environment with MAAS on Z13 systems.

zdev/initramfs/scripts/init-top/s390-tools-zdev
```diff
@@ -56,7 +64,11 @@ done
 
 if [ $zdev_auto -eq 1 ] ; then
 	log_begin_msg "Starting firmware auto-configuration"
-	chzdev --import "$zdev_fw_file" $zdev_base_args
+	if [ -n "$enable_all" ]; then
+		chzdev --all --enable
```
oberpar: Be sure to pass $zdev_base_args to chzdev here or the call might block or fail in other ways.

zdev/initramfs/scripts/init-top/s390-tools-zdev
```diff
@@ -36,8 +36,16 @@ esac
...
 zdev_base_args="--force --yes --no-root-update --no-settle --auto-conf --quiet"
+eval $(grep machine /proc/cpuinfo | cut -f2- -d: | tr -d ' ' | tr ',' ';')
 
-if [ -e "$zdev_fw_file" ] ; then
+case "$machine" in
+# Z13 or Z13s
+2964|2965)
+	enable_all="true"
+	;;
+esac
```
oberpar: I don't think that automatic configuration of all devices should be enabled per default on all z13 machines, as it can have undesired side-effects. Also the functionality is unrelated to z13.

A better approach would be: provide an option for users to specify that they want auto-configuration, even on pre-z14:

1. Extend `rd.zdev` parameter parsing to accept `auto` as value. If specified this sets `zdev_auto=1`
2. If `zdev_auto=1` is set, depending on the availability of the `$zdev_fw_file` import the file, or use `--enable --all`

Also this change needs to be mirrored to `s390-tools/zdev/dracut/95zdev/parse-zdev.sh` (the dracut implementation of the same logic).

hoeppnerj:
Hi @ltrager , any plans on sending an updated version? Thanks!

hoeppnerj:
ping @ltrager 
Do you still plan to send a new version? Otherwise we might close this PR without further processing.

ltrager:
Sorry its taking so long to respond. I lost access to the Z13 system I had. I should be getting access again in the next few weeks. Once I do I'll take another look at this branch and test changes.

xnox:
> Sorry its taking so long to respond. I lost access to the Z13 system I had. I should be getting access again in the next few weeks. Once I do I'll take another look at this branch and test changes.

you should still have access to the s390x LPAR and KVM internally, and z14  via external connection. Maybe you can chat to bjorn or frank heimes to double check?

hoeppnerj:
@ltrager @xnox any updates here?

ltrager:
Sorry its taking so long to get to this. I'm verifying changing now and should have an update next week.

ltrager:
I'm at a bit of a loss as to what I'm doing wrong. I updated the [s390-tools-zdev](https://github.com/ibm-s390-linux/s390-tools/files/6013394/s390-tools-zdev.txt) as suggested. Once I do that the initrd hangs after running `chzdev`. As seen in the [console log](https://github.com/ibm-s390-linux/s390-tools/files/6013397/failure.log) DHCP doesn't get a response now. If I revert back to the version which is currently in this branch everything works fine.

ltrager:
It looks like the Z14 I have access to now doesn't need this patch. `/sys/firmware/sclp_sd/config/data` is found in the initrd and everything loads automatically now :)

Issue #55, opened 2019-02-21
zkey crypttab generates invalid entries for plain type devices


xnox:
The correct syntax as implemented by systemd-cryptsetup and by debian's crypttab package crypttab process is ',plain' or ',luks2' etc to specify the type. 'hash=plain' is invalid for a plain type entry.

ifranzki:
I understand about 'hash=plain' being invalid (I will fix this), but I don't understand about ',plain' or ',luks2'. Why the comma in front of plain or luks? 

The generated crypttab entry currently looks as follows:
`enc-disk        /dev/dasda1      /etc/zkey/repository/plain.skey plain,cipher=paes-cbc-plain64,size=512,hash=plain`

So the 'plain' keyword is as first in the options, so I don't think that there should be a comma in front of it.

xnox:
> I understand about 'hash=plain' being invalid (I will fix this), but I don't understand about ',plain' or ',luks2'. Why the comma in front of plain or luks?
> 

I was trying to point out that it is just a "key" by itself, not a "key=value" parameter. as in 'plain' by itself is the correct syntax. Sorry for confusion.

> The generated crypttab entry currently looks as follows:
> `enc-disk /dev/dasda1 /etc/zkey/repository/plain.skey plain,cipher=paes-cbc-plain64,size=512,hash=plain`
> 

So systemd-cryptsetup i believe does choke up on the ',hash=plain' bit and it should be removed. Otherwise it is valid way to declare the type the "plain," start of it.

However, on Ubuntu, we do not normally specify devices as `/dev/kernelname` instead we use UUID entries, eg. things like this:

```
nvme0n1p7_crypt UUID=a228a613-c3d2-42ec-8799-2b637afc4dce none luks,discard
```

is the structure of entries that I would like `cryptsetup crypttab` generate. ie. Use `UUID=` and always specify `discard` option.

> So the 'plain' keyword is as first in  the options, so I don't think that there should be a comma in front of it.

yes yes

xnox:
> And is it really 'luks2' for a LUKS2 volume, or just 'luks' ? The man page of crypttab specifies only 'luks'.
> I would assume that it finds out by itself it is luks1 or luks2.

I believe the crypttab parser in Debian/Ubuntu cryptsetup package for initramfs-tools unlocking wants a specifier. You are correct that cryptsetup can work out if things are luks1 or luks2 based on luks keyword, but i wonder if we want to be more specific or not. Let me get back on that.

[...]

PR #57, opened 2019-03-06
Drop invalid hash=plain from zkey crypttab plain entries.

xnox:
As discussed with @ifranzki on https://github.com/ibm-s390-tools/s390-tools/issues/55


ifranzki:
I have already a patch that fixes this (and also other stuff) in the pipe that closes issue #55, the patch just needs to be integrated. I will make sure that it gets integrated ASAP.

PR #59, opened 2019-03-07
Passphrase options

xnox:
This adds the common passphrase options to `zkey cryptsetup` subcommand, as well as adds the `-batch-mode` option.

Unfortunately, I was not able to reuse the args with the `zkey-cryptsetup` command, due to `-l` clash... it's both `--volumes` and `--keyfile-size`, hence made it use `-L` short arg.

Fixes https://github.com/ibm-s390-tools/s390-tools/issues/58

### DIFF
--- a/zkey/keystore.c
+++ b/zkey/keystore.c
@@ -3242,6 +3242,10 @@ static int _keystore_execute_cmd(const char *cmd,
 
 struct crypt_info {
 	bool execute;
+	const char *keyfile;
+	size_t keyfile_offset;
+	size_t keyfile_size;
+	size_t tries;
 	char **volume_filter;
 	int (*process_func)(struct keystore *keystore,
 			    const char *volume,
@@ -3283,8 +3287,21 @@ static int _keystore_process_cryptsetup(struct keystore *keystore,
 	char temp[100];
 	int rc = 0;
 	char *cmd;
+	char *common_passphrase_options;
+	size_t common_len;
 
 	sprintf(temp, "--sector-size %lu ", sector_size);
+
+	util_asprintf(&common_passphrase_options, "");
+	if (info->keyfile)
+		util_asprintf(&common_passphrase_options, "%s--key-file %s ", common_passphrase_options, info->keyfile);
+	if (info->keyfile_offset > 0)
+		util_asprintf(&common_passphrase_options, "%s--keyfile-offset %lu ", common_passphrase_options, info->keyfile_offset);
+	if (info->keyfile_size > 0)
+		util_asprintf(&common_passphrase_options, "%s--keyfile-size %lu ", common_passphrase_options, info->keyfile_size);
+	if (info->tries > 0)
+		util_asprintf(&common_passphrase_options, "%s--tries %lu ", common_passphrase_options, info->tries);
+	common_len = strlen(common_passphrase_options);
 
 	if (strcasecmp(volume_type, VOLUME_TYPE_PLAIN) == 0) {
 		util_asprintf(&cmd,
@@ -3310,9 +3327,10 @@ static int _keystore_process_cryptsetup(struct keystore *keystore,
 		util_asprintf(&cmd,
 			      "cryptsetup luksFormat %s--type luks2 "
 			      "--master-key-file '%s' --key-size %lu "
-			      "--cipher %s --pbkdf pbkdf2 %s%s",
+			      "--cipher %s --pbkdf pbkdf2 %s%s%s",
 			      keystore->verbose ? "-v " : "", key_file_name,
 			      key_file_size * 8, cipher_spec,
+			      common_len > 0 ? common_passphrase_options : "",
 			      sector_size > 0 ? temp : "", volume);
 
 		if (info->execute) {
@@ -3327,8 +3345,9 @@ static int _keystore_process_cryptsetup(struct keystore *keystore,
 			return rc;
 
 		util_asprintf(&cmd,
-			      "zkey-cryptsetup setvp %s%s", volume,
-			      keystore->verbose ? " -V " : "");
+			      "zkey-cryptsetup setvp %s %s%s", volume,
+			      common_len > 0 ? common_passphrase_options : "",
+			      keystore->verbose ? "-V" : "");
 
 		if (info->execute) {
 			printf("Executing: %s\n", cmd);
@@ -3581,7 +3600,8 @@ out:
  * @returns 0 for success or a negative errno in case of an error
  */
 int keystore_cryptsetup(struct keystore *keystore, const char *volume_filter,
-			bool execute, const char *volume_type)
+			bool execute, const char *volume_type, const char *keyfile,
+	                size_t keyfile_offset, size_t keyfile_size, size_t tries)
 {
 	struct crypt_info info = { 0 };
 	int rc;
@@ -3598,6 +3618,10 @@ int keystore_cryptsetup(struct keystore *keystore, const char *volume_filter,
 	}
 
 	info.execute = execute;
+	info.keyfile = keyfile;
+	info.keyfile_offset = keyfile_offset;
+	info.keyfile_size = keyfile_size;
+	info.tries = tries;
 	info.volume_filter = str_list_split(volume_filter);
 	info.process_func = _keystore_process_cryptsetup;
 
--- a/zkey/keystore.c
+++ b/zkey/keystore.c
@@ -3242,6 +3242,7 @@ static int _keystore_execute_cmd(const char *cmd,
 
 struct crypt_info {
 	bool execute;
+	bool batch_mode;
 	const char *keyfile;
 	size_t keyfile_offset;
 	size_t keyfile_size;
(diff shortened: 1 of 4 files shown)


zkey/keystore.c
```diff
@@ -3283,9 +3287,22 @@ static int _keystore_process_cryptsetup(struct keystore *keystore,
...
+	if (info->keyfile)
+		util_asprintf(&common_passphrase_options, "%s--key-file %s ", common_passphrase_options, info->keyfile);
+	if (info->keyfile_offset > 0)
+		util_asprintf(&common_passphrase_options, "%s--keyfile-offset %lu ", common_passphrase_options, info->keyfile_offset);
+	if (info->keyfile_size > 0)
+		util_asprintf(&common_passphrase_options, "%s--keyfile-size %lu ", common_passphrase_options, info->keyfile_size);
+	if (info->tries > 0)
+		util_asprintf(&common_passphrase_options, "%s--tries %lu ", common_passphrase_options, info->tries);
+	common_len = strlen(common_passphrase_options);
+
```
ifranzki: You need to free common_passphrase_options at the end of the routine, just like cmd is freed.

zkey/keystore.c
```diff
@@ -3283,9 +3287,22 @@ static int _keystore_process_cryptsetup(struct keystore *keystore,
...
 	int rc = 0;
 	char *cmd;
+	char *common_passphrase_options;
+	size_t common_len;
 
 	sprintf(temp, "--sector-size %lu ", sector_size);
 
+	util_asprintf(&common_passphrase_options, "");
+	if (info->keyfile)
+		util_asprintf(&common_passphrase_options, "%s--key-file %s ", common_passphrase_options, info->keyfile);
```
ifranzki: Please make sure the lines are not over 80 characters

zkey/keystore.c
```diff
@@ -3581,7 +3600,8 @@ static int _keystore_process_crypt(struct keystore *keystore,
  * @returns 0 for success or a negative errno in case of an error
  */
 int keystore_cryptsetup(struct keystore *keystore, const char *volume_filter,
-			bool execute, const char *volume_type)
+			bool execute, const char *volume_type, const char *keyfile,
+	                size_t keyfile_offset, size_t keyfile_size, size_t tries)
```
ifranzki: Please use tabs not spaces to indent.

zkey/keystore.h
```diff
@@ -68,7 +68,8 @@ int keystore_list_keys(struct keystore *keystore, const char *name_filter,
 		       const char *volume_type);
 
 int keystore_cryptsetup(struct keystore *keystore, const char *volume_filter,
-			bool execute, const char *volume_type);
+			bool execute, const char *volume_type, const char *keyfile,
+	                size_t keyfile_offset, size_t keyfile_size, size_t tries);
```
ifranzki: Please use tabs not spaces to indent

zkey/zkey.c
```diff
@@ -1380,7 +1411,7 @@ static int command_cryptsetup(void)
 {
 	int rc;
 
-	rc = keystore_cryptsetup(g.keystore, g.volumes, g.run, g.volume_type);
+	rc = keystore_cryptsetup(g.keystore, g.volumes, g.run, g.volume_type, g.keyfile, g.keyfile_offset, g.keyfile_size, g.tries);
```
ifranzki: Make sure the lines are not over 80 characters

zkey/zkey.c
```diff
@@ -1523,7 +1554,19 @@ int main(int argc, char *argv[])
 			g.tonew = 1;
 			break;
 		case 'o':
-			g.fromold = 1;
+			if (command && !strcmp(command->command, COMMAND_CRYPTSETUP)) {
```
ifranzki: Not sure if I like to reuse the short args of existing options...
How about -O instead?

zkey/zkey.c
```diff
@@ -1539,6 +1582,7 @@ int main(int argc, char *argv[])
 			break;
 		case 'd':
 			g.description = optarg;
+			g.keyfile = optarg;
```
ifranzki: Not sure if I like to reuse the short args of existing options...
How about -D instead?

zkey/zkey.c
```diff
@@ -582,6 +586,33 @@ static struct util_opt opt_vec[] = {
 		.desc = "Runs the generated cryptsetup command",
 		.command = COMMAND_CRYPTSETUP,
 	},
+	{
+		.option = {"key-file", required_argument, NULL, 'd'},
```
ifranzki: Not sure if I like to reuse the short args of existing options...
How about -O instead?

zkey/zkey.c
```diff
@@ -582,6 +586,33 @@ static struct util_opt opt_vec[] = {
...
+		.argument = "FILE-NAME",
+		.desc = "Read the passphrase from the specified file",
+		.command = COMMAND_CRYPTSETUP,
+	},
+	{
+		.option = {"keyfile-offset", required_argument, NULL, 'o'},
```
ifranzki: Not sure if I like to reuse the short args of existing options...
How about -D instead?

zkey/keystore.h
```diff
@@ -68,7 +68,7 @@ int keystore_list_keys(struct keystore *keystore, const char *name_filter,
 		       const char *volume_type);
 
 int keystore_cryptsetup(struct keystore *keystore, const char *volume_filter,
-			bool execute, const char *volume_type, const char *keyfile,
+			bool execute, bool batch_mode, const char *volume_type, const char *keyfile,
```
ifranzki: Make sure the line is not longer than 80 charatcers

zkey/keystore.c
```diff
@@ -3600,7 +3603,7 @@ static int _keystore_process_crypt(struct keystore *keystore,
  * @returns 0 for success or a negative errno in case of an error
  */
 int keystore_cryptsetup(struct keystore *keystore, const char *volume_filter,
-			bool execute, const char *volume_type, const char *keyfile,
+			bool execute, bool batch_mode, const char *volume_type, const char *keyfile,
```
ifranzki: Make sure the line is not longer than 80 charatcers

zkey/zkey.1
```diff
@@ -547,6 +547,15 @@ option to generate crypttab entries for the specified volume type only.
 .RB [ \-\-volume-type | \-t
 .IR type ]
 .RB [ \-\-run | \-r ]
+.RB [ \-\-key\-file | \-d
```
ifranzki: See comments about reusing of short args in the other commits and adapt accordingly.

xnox:
The command line options mimicked here, are those of cryptsetup / zkey-cryptsetup, as the options (both short and long) are the ones used by the cryptsetup tool.

Ideally, I would not want to introduce short options at all, and use long options only, but that's not really an option I don't think. If it is, my preference would be to force long args only.

I find it unfortunate that, zkey cryptsetup and zkey crypttab subcommands are part of zkey tool, rather than the zkey-cryptsetup tool. (I do understand that currently zkey-cryptsetup tool compilation is conditional, but it could compile cryptsetup/crypttab subcommands always).

It is annoying that '-l' option is already in use by the `zkey cryptsetup` hence I cannot really use `-l` again to specify the --keyfile-size. I guess I could enforce positional subflag parsing, but using '-L' seemed like the least work/surprise.

I also do not understand the reuse of short args concern raised, given that cryptsetup subcommand does not use the other flags - and unlikely to start doing so in the future. Enforcing short args uniqueness across the whole tool, is desirable, but counter-intuitive when cryptsetup itself will be called.

I understand that using same short option with different meanings, depending on which subcommand is doing parsing is awkward, but imho it is ok here, as imho cryptsetup well-known args should take precedence.

To be honest, I'd rather move the 'crypsetup|crypttab' subcommands into 'zkey-cryptsetup' and use the `OPT_PASSPHRASE_ENTRY(cmd)` from there, which does use cryptsetup compatible long and short options.

ifranzki:
> Ideally, I would not want to introduce short options at all, and use long options only, but that's not really an option I don't think. If it is, my preference would be to force long args only.

Yes I guess that would be me preference, too, and this is possible:



```
#define OPT_KEYFILE	256    // <-- use values in the range above 256 for different options....

	{
		.option = {"key-file", required_argument, NULL, OPT_KEYFILE},
		.argument = "FILE-NAME",
		.desc = "Read the passphrase from the specified file",
		.command = COMMAND_CRYPTSETUP,
		.flags = UTIL_OPT_FLAG_NOSHORT,
	},
```
In the switch you check for OPT_KEYFILE instead of the short option character. 


> I find it unfortunate that, zkey cryptsetup and zkey crypttab subcommands are part of zkey tool, rather than the zkey-cryptsetup tool.

> To be honest, I'd rather move the 'crypsetup|crypttab' subcommands into 'zkey-cryptsetup' and use the OPT_PASSPHRASE_ENTRY(cmd) from there, which does use cryptsetup compatible long and short options.

One reason is that zkey operates on the key repository, which has all the uinformation about the to be generated commands and crypttab entriers. zkey-cryptsetup does not know the repository at all, and just acts on LUKS2 volumes. I would like to keep that separation.

ifranzki:
In the meantime I took your patches and adjusted them to meet my requirements. I also found a few additional small things that needed to be changed.

Are you OK when I push the modified patches and add you as `Suggested-by: Dimitri John Ledkov <xnox@ubuntu.com>` to the commit message?

xnox:
@ifranzki sure

PR #61, opened 2019-05-02
zipl: update stage3 objcopy command for gcc9

sharkcz:
The objcopy command for stage3.bin needs to take the new rodata section into account.
See also PR #60.

Signed-off-by: Dan Horák <dan@danny.cz>


hoeppnerj:
Thanks for the contribution. Already integrated internally.

PR #63, opened 2019-06-19
Ftbfs stage3

xnox:
zipl/boot/Makefile: fix failure to build, do not specify .lds twice.
    
Whilst stage3.lds is a dependency, it is specified via SFLAGS and must not be repeated again along all the .o files. Filter anything but .o files.

Fixes failure to build from source on Ubuntu 19.10.
    
Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/s390-tools/+bug/1833238


stefan-haberland:
Thanks for the patches. They look good to me and will be added asap.

Issue #70, opened 2019-09-18
zipl: options to allow BLS files only, aka no zipl.conf


tuan-hoang1:
Hi Stefan @stefan-haberland,

If BLS conf files are used, a `/etc/zipl.conf` might be as simple as below from Fedora 30:
```
[defaultboot]
defaultauto
prompt=1
timeout=5
target=/boot
```
The code that generated this zipl.conf file is at : https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/bootloader/zipl.py#L132

@martinezjavier suggested if we could have a way to ***not*** to have this simple zipl.conf file existed.

My suggestion is to have a cmdline option like`--dummy-conf` or `--simple-conf`, which assume above options without generating an actual `/etc/zipl.conf` file (please correct me Javier if I'm wrong here).

tuan-hoang1:
Quick talk with Stefan today, we decided I would create a patch for this feature. Doing now.

lucab:
Here is an alternative suggestion, which fits in the CoreOS model and avoid hardcoding configuration in code. I don't know a lot about zipl though, so feel free to adapt to your case.

The lookup logic for `zipl.conf` could be augmented to fall through these directories, in order of priority, and select the existing path with the highest priority:
 1. `/run/zipl/zipl.conf`
 2. `/etc/zipl.conf`
 3. `/usr/lib/zipl/zipl.conf`

In the normal case, this won't change the current behavior, and it will keep reading `/etc/zipl.conf`.
However distributions can now start shipping a `/usr/lib/zipl/zipl.conf` to be used as a default if (and only if) the file under `/etc` does not exist.

This allows distributions to ship (and update) the default settings, while allowing users to keep their customizations under `/etc` (persistent) and `/run` (volatile, only for the current boot).

(Path naming is just an example, feel free to pick what fits for you)

sharkcz:
@tuan-hoang1, please also set the secure=auto in the default config.

sharkcz:
Using `/lib/s390-tools/zipl.conf` as default is good idea too.

tuan-hoang1:
@stefan-haberland , @martinezjavier : Sorry, was preempted by other tasks ... How does the very simple patch below look ?:
```
diff --git a/zipl/Makefile b/zipl/Makefile
index a02e997..931b113 100644
--- a/zipl/Makefile
+++ b/zipl/Makefile
@@ -8,6 +8,7 @@ all:
 install: all
[...]

PR #71, opened 2019-09-30
zipl: check multiple locations of zipl.conf, ship a minimal zipl.conf

tuan-hoang1:
Fixes: #70
https://github.com/ibm-s390-tools/s390-tools/issues/70


zipl/include/zipl.h
```diff
@@ -47,7 +47,9 @@
 #define DEFAULTBOOT_SECTION		"defaultboot"
 
 #define ZIPL_CONF_VAR			"ZIPLCONF"
+#define ZIPL_RUNTIME_CONF		"/run/zipl.conf"
```
lucab: I think it is customary to use a dedicated sub-directory for this (similar to `libdir` below here): `/run/s390-tools/zipl.conf`.

martinezjavier: Agreed. Probably `/run/zipl/zipl.conf` since the configuration is for the `zipl` tool and `s390-tools` seems to be more of a repository that groups different tools.

zipl/man/zipl.conf.5
```diff
@@ -17,8 +17,13 @@ boot loader tool
 .BR zipl (8)).
 .br
 
-By default this configuration file is located at /etc/zipl.conf. A different
-location may be specified either using the '\-\-config' option of
+By default
+.B zipl
+checks for
+.I zipl.conf
+at /run/zipl.conf, /etc/zipl.conf, /lib/s390-tools/zipl.conf in that order -
```
martinezjavier: The `/lib/s390-tools/zipl.conf` will depend on the value set to `TOOLS_LIBDIR` right? Or is this always set to `$(INSTALLDIR)/lib/s390-tools`? I see that `/etc` is also the value set to `TOOLS_SYSCONFDIR`, but if I'm reading the code correctly this is always set to `$(INSTALLDIR)/etc`.

I'm asking in case a user ends having the minimal config file for example in `/usr/lib/s390-tools/zipl.conf` and getting confused since the man page doesn't match their setup.

tuan-hoang1: I actually only care about `DESTDIR` for packaging and users who get zipl/s390-tools from distro.

I thought those who use `INSTALLDIR` know what they are doing (temporarily testing a component of s390-tools, etc.). Should we make the assumption that they should know that their minimal zipl.conf is now at, for example, `/tmp/lib/s390-tools/zipl.conf` if they built with `INSTALLDIR=/tmp make install`, intead of `/lib/s390-tools/zipl.conf` ? Now if we want to cover those users who compile themselves by using `INSTALLDIR`, well, let's just use

`#define ZIPL_MINIMAL_CONF		"/lib/s390-tools/zipl.conf"`

but I'm not sure this is something we should do.

zipl/src/job.c
```diff
@@ -1797,7 +1797,13 @@ get_job_from_config_file(struct command_line* cmdline, struct job_data* job)
...
 	} else {
 		/* Use default config file */
-		filename = ZIPL_DEFAULT_CONF;
+		if (misc_check_readable_file(ZIPL_RUNTIME_CONF) == 0) {
+			filename = ZIPL_RUNTIME_CONF;
+		} else if (misc_check_readable_file(ZIPL_DEFAULT_CONF) == 0) {
+			filename = ZIPL_DEFAULT_CONF;
+		} else if (misc_check_readable_file(ZIPL_MINIMAL_CONF) == 0) {
+			filename = ZIPL_MINIMAL_CONF;
+		}
```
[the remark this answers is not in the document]
tuan-hoang1: Agreed.

zipl/include/zipl.h
tuan-hoang1: Sure. Since I'm pretty new to this `/run` custom/convention, would you mind copying some code where this custom being used. Just being curious and want to learn.

zipl/man/zipl.conf.5
martinezjavier: I think what we want is to have a `zipl/man/zipl.8.in` file that's a template and have `TOOLS_SYSCONFDIR` and the others variables and a `sed` command or something in the makefile that generates the `zipl/man/zipl.8` file that's installed.

That way the man page will reflect the values that were used to build the tool.

tuan-hoang1:
The last 2 force pushes were for fixing man pages.

zipl/src/job.c
```diff
@@ -59,6 +59,14 @@ static struct option options[] = {
 /* Command line option abbreviations */
 static const char option_string[] = "-c:b:t:i:r:p:P:d:D:M:s:S:m:hHnVvaT:fk:";
 
+/* Locations of zipl.conf configuration file */
+static const char *zipl_conf[] = {
+	ZIPL_RUNTIME_CONF,
+	ZIPL_DEFAULT_CONF,
+	ZIPL_MINIMAL_CONF,
+	NULL
```
lucab: I _think_ you don't strictly need this. You can statically get the number of entries in this array as `sizeof(array)/sizeof(entry)`.

tuan-hoang1: Yes, there exists `ARRAY_SIZE` macro in `zipl.h` to do this task. Both of this and that should just work, there are places in s390-tools where they are both used.

stefan-haberland:
Hi,

thanks a lot for the cotribution. From reading the Code it looks good to me. I will give it a test and include it if there is no additional finding.

Regards,
Stefan

Issue #69, opened 2019-09-18
zipl: relative path for BLS values


tuan-hoang1:
Hi Javier @martinezjavier and Stefan @stefan-haberland,

Currently, zipl assumes the kernel and ramdisk paths to be absolute, both in zipl.conf and BLS conf files, or else it will print out file not found error to user.

In Fedora/RHEL CoreOS, a BLS file might look like:
```
initrd /ostree/fedora-coreos-<commit-id>/initramfs-5.2.11-200.fc30.x86_64.img
linux /ostree/fedora-coreos-<commit-id>/vmlinuz-5.2.11-200.fc30.x86_64
```
which conforms with the BLS spec, as it says the `linux` and `initrd` fields `shall be a path relative to the $BOOT directory`.

I suggest we support this case, by also checking the paths of the images with a prefix of `/boot`, and only error out when both cases (with and without `/boot` path prefix) cannot find the files. Below is a patch I would like propose: (https://tpaste.us/yPMy)

```
diff --git a/zipl/src/scan.c b/zipl/src/scan.c
index b42b5f2..3329005 100644
--- a/zipl/src/scan.c
+++ b/zipl/src/scan.c
@@ -672,7 +672,7 @@ static int
 scan_bls_field(struct misc_file_buffer *file, struct scan_token* scan,
 	       int* index)
 {
-	int current;
+	int current, rc;
 	int key_start, key_end;
 	int val_start, val_end;
 	char *val;
@@ -712,6 +712,15 @@ scan_bls_field(struct misc_file_buffer *file, struct scan_token* scan,
 
 	if (strncmp("linux", &file->buffer[key_start], key_end - key_start) == 0) {
 		misc_asprintf(&val, "%s", &file->buffer[val_start]);
+		rc = misc_check_readable_file(val);
+		if (rc) {
+			misc_asprintf(&val, "/boot%s", &file->buffer[val_start]);
+			rc = misc_check_readable_file(val);
+			if (rc) {
+				error_text("Image file '%s'", val);
+				return rc;
+			}
+		}
 		scan_append_keyword_assignment(scan, index, scan_keyword_image, val);
 		free(val);
 	}
@@ -723,6 +732,15 @@ scan_bls_field(struct misc_file_buffer *file, struct scan_token* scan,
 
 	if (strncmp("initrd", &file->buffer[key_start], key_end - key_start) == 0) {
 		misc_asprintf(&val, "%s", &file->buffer[val_start]);
[...]

PR #74, opened 2019-10-04
zipl: add value of target= as search path for BLS case

tuan-hoang1:
In the BLS files, the 'linux' and 'initrd' fields are relative to the
/boot directory. If /boot is a mount point of boot partition, 'linux'
and 'initrd' fields are relative to boot partition. If not, the paths
are relative to /boot directory of root partition.

zipl always starts searching at / regardless of partitions and mount
points.

This commit extends the semantics of target= field in zipl.conf to make
it as a second search path, besides /, for those 2 BLS fields.

See issues #69 for more details.

Fixes: #69

Signed-off-by: Tuan Hoang <tmhoang@linux.ibm.com>


tuan-hoang1:
Force pushes are to fix commit message, no functional change.

stefan-haberland:
Hi,

thanks a lot for the contribution. From reading the Code it looks good to me. I will give it a test and include it if there is no additional finding.

Regards,
Stefan

PR #79, opened 2020-01-22
zdev/dracut/95zdev/module-setup.sh: Add ctcm kernel module

markkp:
The module-setup.sh script doesn't include the ctcm module
when building the initrd. This causes any CTC devices to
be bound by the LCS driver instead.

Fix for bsc#1160373.

Signed-off-by: Mark Post <mpost@suse.com>


hoeppnerj:
Thanks, integrated into our internal repo.

PR #87, opened 2020-06-03
lscpumf: add z15 Model T02

sharkcz:


sharkcz:
Machine type `8562` is handled in `get_counter()`, but missing in `machine_name()`, so be consistent.

sharkcz:
ping :-)

cpumf/lscpumf.c
```diff
@@ -2572,6 +2572,7 @@ static const char *machine_name(void)
 	case 3906:	return "IBM z14";
 	case 3907:	return "IBM z14 ZR1";
 	case 8561:	return "IBM z15";
+	case 8562:	return "IBM z15 Model T02";
```
hbrueckner, line 2575: Hi Dan,

I just checked the current branding guidelines and it should be "IBM z15; Model T02" which is now a bit different compared to the z14 naming.

hbrueckner (commented):
> ping :-)

pong :)

I just added a short remark on the naming. Thomas is currently out and I would ask him for a final review when he is back.

cpumf/lscpumf.c
sharkcz: thanks, PR updated, there might be other places/projects that would need an update of the branding

tmricht:
@sharkcz Dan, can you please omit the semicolon after z15.
As hbrückner posted in his append.

sharkcz:
done, I thought he wants me to add the semicolon :-)

PR #94, opened 2020-10-28
add simple man page for the mk-s390image script

sharkcz:


sharkcz:
We might end up with making mk-s390image a regular tool if things go well in https://bugzilla.redhat.com/show_bug.cgi?id=1891778. I would make a new PR for that.

mihajlov:
Would be fine with me, let's see what @hoeppnerj has to say.

hoeppnerj:
Seems fine to me, I'll pull it

PR #95, opened 2020-10-28
Clear kernel cmdline data before writing new parmfile

bcl:
If this isn't cleared first it can end up with an unexpected cmdline if
it doesn't completely overwrite the default data.


mihajlov:
Makes sense and looks good to me.

mihajlov (approved):
Looks good to me.

hoeppnerj:
Integrated internally. Thanks for your contribution!

PR #106, opened 2021-01-07
zipl/boot/Makefile: -no-pie is not a valid ld flag

slyon:
The "-no-pie" option never has been a valid flag of ld. It breaks the build with newer binutils.
See:
https://sourceware.org/bugzilla/show_bug.cgi?id=27050
https://bugs.launchpad.net/ubuntu/+source/s390-tools/+bug/1907789

### DIFF
--- a/zipl/boot/Makefile
+++ b/zipl/boot/Makefile
@@ -109,7 +109,7 @@ stage3.bin:	stage3.exec
 		$< $@
 
 data.o: $(FILES)
-	$(LD) $(NO_PIE_LDFLAGS) -r -b binary -o data.o $(FILES)
+	$(LD) -r -b binary -o data.o $(FILES)
 
 data.h: data.o
 	rm -f data.h


sharkcz:
Shouldn't all occurrences of the NO_PIE_LDFLAGS variable go away then? I mean remove it also from common.mak and review the use of NO_PIE_LINKFLAGS vs NO_PIE_LDFLAGS

slyon:
Yes. The NO_PIE_LINKFLAGS vs NO_PIE_LDFLAGS should be reviewed, as their differentiation was not really clear to me.
But I saw that in some places the NO_PIE_LDFLAGS are also used as a compiler flag, where it is valid and needed. It's just invalid as a `ld` flag.

As an external contributor I did not want do fiddle too much with the build system, but would rather leave this to somebody more involved with the project. This PR is just the smallest possible fix I could find and applied to the Ubuntu package.

xnox:
I thought that `-no-pie` is valid for some tooling, i.e. `gcc` accepts it and does things with it? but not `ld`?

slyon:
> `gcc` accepts it and does things with it? but not `ld`?

Exactly.

hoeppnerj:
So, using -no-pie for ld "worked" just by chance because ld was bugged and didn't properly check what was specified? Am I getting that right?

Anyway, the PR seems to be okay. Looking into the other NO_PIE_* stuff as well though before pulling it. Thanks for the hint!

slyon:
Correct @hoeppnerj
`ld` just ignored that flag earlier, but now it complains about it (as of binutils >= 3.35.50). And upstream tells us: "The linker does not have a -no-pie or a --no-pie command line option, so the error message is correct."

Thanks for looking into it!

hoeppnerj:
@slyon I've prepared a patch internally to get rid of the NO_PIE_LDFLAGS entirely. I'll merge your patch with that and add you as Suggested-by, would that be okay with you?

sharkcz:
a friendly ping :-) It's breaking builds in Fedora Rawhide with binutils 2.36.1-15.fc35

`ld: Error: unable to disambiguate: -no-pie (did you mean --no-pie ?)`

hoeppnerj:
@slyon sorry, it's been a while. We have a version internally now that fixes the build issues.
However, instead of removing the `-no-pie` option we instead use GCC for the linking as it should be
done anyway. This particular part was the only one that used `ld` directly.
With that we fix two things in one patch.

That also means that I won't take your patch here. But I would gladly mention you as Suggested-by and
reference this PR if that's okay for you.

slyon:
Sounds good to me, thank you!

PR #107, opened 2021-01-27
zipl: fix reading 4k disk's geometry

nikita-dubrovskii:
Fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1918723

On 4k SCSI disks zipl stores wrong values to 'scsi_mbr.program_table_pointer',
which makes system unbootable.
This happens in 'zipl/src/disk.c:656':

```
/* Convert file system block to physical */
*physical = mapped * phy_per_fs + subblock;
/* Add partition start */
*physical += info->geo.start;

```

So 'hd_geometry.start' should be adjusted before being used.

Signed-off-by: Nikita Dubrovskii <nikita@linux.ibm.com>

### DIFF
--- a/zipl/src/disk.c
+++ b/zipl/src/disk.c
@@ -444,6 +444,9 @@ type_determined:
 	}
 	/* Convert device size to size in physical blocks */
 	data->phy_blocks = devsize / (data->phy_block_size / 512);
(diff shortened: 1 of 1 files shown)


stefan-haberland (approved):

stefan-haberland:
Thanks for the patch. With the rework I am fine with it.

Issue #111, opened 2021-05-17
Cannot use the Menu feature with BLS entries


rmetrich:
Trying to create a menu sourcing one of more BLS entries, **zipl** fails to generate the configuration and reports this:
~~~
# zipl -V -b . -c zipl.conf.menu --dry-run
Using config file 'zipl.conf.menu' (from command line)
Using BLS config file './bls1.conf'
Error: Config file 'zipl.conf.menu': Line 0: missing keyword 'target' in section 'bls1
~~~

The official documentation https://www.ibm.com/docs/en/linux-on-systems?topic=loader-bls-files has been followed.

### bls1.conf

~~~
title bls1
version 4.18.0-240.el8.s390x
linux /boot/vmlinuz-4.18.0-240.el8.s390x
initrd /boot/initramfs-4.18.0-240.el8.s390x.img
~~~

### zipl.conf.menu

~~~
[defaultboot]
defaultmenu=blsmenu

:blsmenu
1="bls1"
prompt=1
timeout=5
secure=auto
target=/boot
~~~

## Source code (scan.c)

~~~
1035         /* Check keywords */
1036         for (i=0; i < SCAN_KEYWORD_NUM; i++) {
1037                 switch (scan_key_table[(int) *type][i]) {
1038                 case req:
1039                         /* Check for missing data */
1040                         if ((keyword[i] == 0) && (line != NULL)) {
1041                                 /* Missing keyword in config file section */
1042                                 error_reason("Line %d: missing keyword '%s' "
1043                                              "in section '%s'", section_line,
1044                                              scan_keyword_name(
1045                                                      (enum scan_keyword_id) i),
1046                                              name);
1047                                 return -1;
~~~

Executing the command under GDB shows that **type** is **section_ipl**, which indeed requires the **target** keyword:
~~~
  45 enum scan_key_state scan_key_table[SCAN_SECTION_NUM][SCAN_KEYWORD_NUM] = {
  46 /*       defa dump dump imag para parm ramd segm targ prom time defa tape mv
  47  *       ult  to   tofs e    mete file isk  ent  et   pt   out  ultm      dump
  48  *                           rs                                 enu
 :
  64 /* ipl          */
  65         {inv, inv, inv, req, opt, opt, opt, inv, req, inv, inv, inv, inv, inv,
  66          opt, opt, opt, opt, opt, inv, opt, opt},
~~~

The command fails as soon as BLS entries are read.

sharkcz:
[...]

PR #113, opened 2021-05-28
zipl: make IPL sections defined with BLS to inherit a target field

martinezjavier:
The target field is required for IPL sections, but the ones populated from
BLS snippets did not have this field. This was only working when using BLS
snippets because IPL sections inherit the target field with "defaultauto".

But that's not the case when using a menu, for example the following zipl
configuration will fail if is used along with BLS defined IPL sections:
fail when running the zipl tool:

  [defaultboot]
  defaultmenu = menu1

  :menu1
  target = /boot
  1 = linux
  2 = test
  default = 1
  prompt = 1
  timeout = 0

Because "linux" and "test" will be defined using BLS snippets and these
don't have an option to define target fields. Let's make these to always
inherit a target field, either from the [defaultboot] section or a menu
section if there's a defaultmenu defined.

Fixes: https://github.com/ibm-s390-linux/s390-tools/issues/111

Reported-by: Renaud Métrich <rmetrich@redhat.com>

### DIFF
--- a/zipl/src/scan.c
+++ b/zipl/src/scan.c
@@ -794,102 +794,6 @@ static int sort_bls_fields(struct misc_file_buffer *file, char *filename)
 	return 0;
 }
 
-int
-scan_bls(const char* blsdir, struct scan_token** token, int scan_size)
-{
-	int count = 0;
-	int size, remaining = 0, n, current, rc = -1;
-	struct scan_token* buffer;
-	struct scan_token* array = *token;
-	struct dirent** bls_entries;
-	struct misc_file_buffer file;
-	struct stat sb;
-	char filename[PATH_MAX];
-
-	if (!(stat(blsdir, &sb) == 0 && S_ISDIR(sb.st_mode)))
-		return 0;
-
-	n = scandir(blsdir, &bls_entries, bls_filter, bls_sort);
-	if (n <= 0)
-		return n;
-
-	while (array[count].id != 0)
-		count++;
-
-	remaining = scan_size - count;
-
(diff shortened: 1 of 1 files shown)


zipl/src/scan.c
```diff
@@ -1358,6 +1262,165 @@ scan_get_section_keywords(struct scan_token* scan, int* index, char* name,
 }
 
 
+/* Get the default target defined either in the defaultboot section or the
+ * menu section for the defaultmenu if there is a default menu defined. */
```
stefan-haberland, line 1266: please put the trailing */ on a separate line for block comments

stefan-haberland (changes requested):
Thanks again for this patch.
I tested the code with some scenarios I can imagine and it worked quite well.

The only thing that does not work (but I think it is intended this way) is when you do not have a defaultmenu= set in the zipl.conf but specify the menu on the commandline using "zipl -m menu1" for example. I am also not sure if this can be achieved that easy.

Beside of this I only have some coding style nitpicks. I know that you just copied a lot of the code code and it looks this way all over but maybe we can fix it for the parts you just touched.
General remark: all those "foo* bar" declarations should rather be "foo *bar". Rest of the comments in the patch. Please let me know if you would like to fix the style or if we should the code as is.

best regards,
Stefan

hoeppnerj:
@martinezjavier ping, what is the status of this PR? Are you going to address the last few change requests?

hoeppnerj:
fyi, as there was no response since last year we've pulled the patches and adapted the requested style changes ourselves. Thanks for the contribution!

PR #118, opened 2021-07-15
hsavmcore: fix the dependency evaluation

sharkcz:
Set the Makefile depedency similarily to cmsfs-fuse, so the overlay.c
won't be recompiled during the "make install" phase.

### DIFF
--- a/hsavmcore/Makefile
+++ b/hsavmcore/Makefile
@@ -56,14 +56,12 @@ objects := $(patsubst %.c,%.o,$(sources))
 
 libs = $(rootdir)/libutil/libutil.a
 
-all: hsavmcore
+all: check-dep hsavmcore
 
 hsavmcore: $(objects) $(libs)
 	$(LINK) $(ALL_LDFLAGS) $^ $(LDLIBS) -o $@
 
-overlay.o: check-dep-fuse overlay.c overlay.h
-
-check-dep-fuse:
+check-dep:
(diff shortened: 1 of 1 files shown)


ifranzki:
check-dep-fuse is a dependency of overlay.o (and not 'all') for a good reason. I one calls make with target 'hsavmcore' only, then no fuse-checking would happen. 
The better solution would be to make check-dep-fuse a non-PHONY target, i.e. add `touch check-dep-fuse` to the end of the receipt of check-dep-fuse. That way it is executed only once (during compile of the code), but not anymore during make install.

sharkcz:
That should likely work too, I am not insisting on my solution :-) But would be great ,if there would be consistency between the various tools/subdirs.

eaibmz:
Thanks for reporting.
Should be fixed in the next release.
I went with the solution proposed by Ingo.

sharkcz:
@eaibmz, do you have the patch available somewhere, so it could be integrated in our package? I would rather go with the upstream version, when it exists.

eaibmz:
> @eaibmz, do you have the patch available somewhere, so it could be integrated in our package? I would rather go with the upstream version, when it exists.

Here: https://github.com/eaibmz/s390-tools/commit/f21ec6f5ee853231c910fc48b8a6ab0f33aa255c

sharkcz:
thanks :-)

PR #121, opened 2021-08-01
Makefile: add {CURL,XML2}_CONFIG

ffontaine:
Add `CURL_CONFIG` and `XML2_CONFIG` to allow the user to configure the binaries used in `check-dep-libkmipclient`

This will allow the user to avoid the following build failures raised since version 2.17.0 and https://github.com/ibm-s390-linux/s390-tools/commit/56fecf1832c3ebc4626ddf5c598762833c362d5e:

```
/bin/sh: 1: curl-config: not found
Error: libcurl is not built with the OpenSSL backend
make[2]: *** [Makefile:54: check-dep-libekmfweb] Error 1
```

Fixes:
 - http://autobuild.buildroot.org/results/cfb46d7547c22f0a23aca2286dc5a1d2c20aadfc

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/libekmfweb/Makefile
+++ b/libekmfweb/Makefile
@@ -47,6 +47,8 @@ detect-openssl-version.dep:
 	echo "}" >> $(TMPFILE)
 	mv $(TMPFILE) $@
 
+CURL_CONFIG ?= curl-config
+
 check-dep-libekmfweb: detect-openssl-version.dep
 	$(call check_dep, \
 		"libekmfweb", \
@@ -64,8 +66,8 @@ check-dep-libekmfweb: detect-openssl-version.dep
 		"curl/curl.h", \
 		"libcurl-devel", \
 		"HAVE_LIBCURL=0" \
-		`curl-config --cflags` `curl-config --libs`)
-	curl-config --ssl-backends | grep OpenSSL >/dev/null 2>&1 || { echo "Error: libcurl is not built with the OpenSSL backend"; exit 1; }
+		`$(CURL_CONFIG) --cflags` `$(CURL_CONFIG) --libs`)
+	$(CURL_CONFIG) --ssl-backends | grep OpenSSL >/dev/null 2>&1 || { echo "Error: libcurl is not built with the OpenSSL backend"; exit 1; }
 	touch check-dep-libekmfweb
 
 skip-libekmfweb-openssl:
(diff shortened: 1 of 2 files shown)


yann-morin-1998:
For consistency, shouldn't we also use `CURL_CONFIG` in `libekmfweb/Makefile` too?

ffontaine:
Thanks @yann-morin-1998, I updated the PR.

ifranzki:
Looks good to me.

sharkcz:
There are other tools requiring eg. curl, so it would be good to have a consistency among them. dasdview, genprotimg and zdsfs all use `pkg-config` for retrieving the compiler and linker flags for the curl library.

ffontaine:
So what is your suggestion? Replace curl-config and xml2-config calls by pkg-config?

sharkcz:
> So what is your suggestion? Replace curl-config and xml2-config calls by pkg-config?

yes, I would standardize around pkg-config

ifranzki:
pkg-config is not always available, that's why these other tools have logic to check if pkg-config is available or not. Example from dasdview:

```
ifneq ($(shell sh -c 'command -v pkg-config'),)
CURL_CFLAGS = $(shell pkg-config --silence-errors --cflags libcurl)
CURL_LDLIBS = $(shell pkg-config --silence-errors --libs libcurl)
else
CURL_CFLAGS = -I/usr/include/s390x-linux-gnu/curl
CURL_LDLIBS = -lcurl
endif # shell
```
Not sure If I really like this..... Makes things more complicated actually....

ffontaine:
Indeed, and there is an other issue, there is no direct equivalent to `curl-config --ssl-backends` through `pkg-config`. We could check if `-lssl` is in `Libs` but it's a bit ugly.

sharkcz:
When isn't `pkg-config` available? I would consider is a distro packaging bug ... I think `pkg-config` exists to replace random `*-config` scripts.

sharkcz:
> Indeed, and there is an other issue, there is no direct equivalent to `curl-config --ssl-backends` through `pkg-config`. We could check if `-lssl` is in `Libs` but it's a bit ugly.

Then the pkgconfig file for curl should be fixed IMO.

PR #122, opened 2021-10-29
zipl/boot: Always clear COMMAND_LINE_EXTRA region

huth:
Always clear the COMMAND_LINE_EXTRA region, also in case sclp_setup()
fails. If the region is not cleared properly, there might be junk
in there after a reboot, causing confusion for the kernel later.


mhartmay (approved):

hoeppnerj:
I've pulled the commit. Thanks for your contribution.

PR #124, opened 2021-11-18
cpictl: fix permissions

sharkcz:
The regular 0755 permission should be sufficient for cpictl.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2024102

Signed-off-by: Dan Horák <dan@danny.cz>


hoeppnerj:
Makes sense, pulled. Thanks!

PR #133, opened 2022-05-26
zipl: be verbose about secure boot state

sharkcz:
When trying to understand what's going on in https://bugzilla.redhat.com/show_bug.cgi?id=2089134 (IBM LTC bz 198327) where the secure boot boot record is not always written, I found it would be good have zipl to report the state of secure boot support on the system.

### DIFF
--- a/zipl/src/bootmap.c
+++ b/zipl/src/bootmap.c
@@ -124,15 +124,26 @@ check_secure_boot_support(void)
 	unsigned int val;
 	FILE *fp;
 
+	if (verbose)
+		printf("Secure boot support: ");
+
 	fp = fopen(ZIPL_SIPL_PATH, "r");
-	if (!fp)
+	if (!fp) {
+		if (verbose)
+			printf("not available\n");
 		return false;
+	}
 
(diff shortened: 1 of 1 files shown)


zipl/src/bootmap.c
```diff
@@ -125,7 +128,7 @@ check_secure_boot_support(void)
 	FILE *fp;
 
 	if (verbose)
-		printf("  Secure boot support: ");
+		printf("Secure boot support: ");
```
stefan-haberland, line 131: This should be part of the first patch.

stefan-haberland:
Overall this looks good to me. Only one thing that should be improved. Thanks.

zipl/src/bootmap.c
sharkcz: OK

sharkcz:
updated and rebased

sharkcz:
Hi @hoeppnerj , could you take a look, please?

hoeppnerj:
Pulled and merged. Thanks a lot @sharkcz !

PR #136, opened 2022-06-03
README.md: Add 70-chreipl-fcp-mpath.rules to the list of udev rule descriptions

frank-heimes:
All udev rules that are part of the s390-tools package are listed in the main
README.md with a brief description.
This commit adds information about the newly added rule for chreipl-fcp-mpath,
called '70-chreipl-fcp-mpath.rules'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ Package contents
    - 40-z90crypt.rules: udev rules for z90crypt driver
    - 90-cpi.rules: udev rule to update Control-Program-Information when KVM is
                    used.
+   - 70-chreipl-fcp-mpath.rules: udev rules to monitor multipath events for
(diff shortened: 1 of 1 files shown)


hoeppnerj:
Makes sense, thanks a lot!

Benjamin-Block:
I don't really mind having the list extended with the udev rules from `chreipl-fcp-mpath`.

So far that list only listed rules in `etc/udev/rules.d`, it also misses rules from `zdev`; and I tried to keep information regarding `chreipl-fcp-mpath` bundled in it's own README.md for maintenance reasons. But if you think it'll help distributions/users I don't think this will add anything worth much in terms of maintenance-burden.

PR #145, opened 2022-12-12
fix Makefile install step chown warnings

ngueorguiev:
Since coreutils v9.1 commit 8f31074cb ("chown: warn about USER.GROUP") chown utility now warns about using of wrong separator for USER and GROUP options.

signed-off: Nikolay Gueorguiev <nikolay.gueorguiev@suse.com>


hoeppnerj:
Makes sense, looks good, and pulled. Thanks!
I've actually missed looking at other tools when I saw https://github.com/ibm-s390-linux/s390-tools/commit/bca8fa3f9c8bc7032763c7c150c60acd63195bc6 and I reckon you took that commit as a hint :)

PR #153, opened 2023-07-12
fix permissions in /usr/include

sharkcz:
Fix directory permissions in /usr/include.

`rpmlint` run against Fedora rpms complains about
````
...
s390utils-devel.s390x: E: non-standard-dir-perm /usr/include/ekmfweb 770
s390utils-devel.s390x: E: non-standard-dir-perm /usr/include/kmipclient 770
...
````


ifranzki (approved):
looks good to me.

hoeppnerj:
Integrated. Thanks @sharkcz for the patches!

PR #154, opened 2023-08-23
chreipl: impove disk type detection when running under QEMU

nikita-dubrovskii:
User can attach disk with smth like: "-device virtio-scsi-ccw,... -device scsi-hd,...", so virtio block device appears as '/dev/sda' instead of '/dev/vda'. chreipl assumes all '/dev/sd*' disks as FCP disks, which is not a case in such setup.

### DIFF
--- a/ipl_tools/ccw.c
+++ b/ipl_tools/ccw.c
@@ -9,13 +9,35 @@
  * it under the terms of the MIT license. See LICENSE for details.
  */
 
+#include <assert.h>
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
+#include "lib/util_path.h"
 #include "ipl_tools.h"
+
+/*
+ * Look up for the device in /sys/devices/ hierarchy.
+ *
+ * path must be PATH_MAX large and the value will be replaced in place
+ */
+static int device_sysfs_path(const char *device, char *path, const size_t path_size)
+{
+	assert(device);
+	assert(path);
+	assert(path_size == PATH_MAX);
+	char *buf = util_path_sysfs("block/%s/device", device);
+
+	if (!realpath(buf, path)) {
+		free(buf);
+		return -1;
+	}
+	free(buf);
+	return 0;
+}
 
 /*
  * Check if the specified device number is a valid device number
@@ -77,11 +99,9 @@ out_fclose:
  */
 static int ccw_busid_get_sysfs_new(const char *device, char *busid)
 {
-	char path[PATH_MAX], buf[4096];
+	char path[PATH_MAX] = { '\0' };
 
-	memset(buf, 0, sizeof(buf));
-	snprintf(path, sizeof(path), "/sys/block/%s/device", device);
-	if (realpath(path, buf) == NULL)
+	if (device_sysfs_path(device, path, sizeof(path)) != 0)
 		return -1;
 
 	/*
@@ -89,7 +109,7 @@ static int ccw_busid_get_sysfs_new(const char *device, char *busid)
 	 * /sys/devices/css0/0.0.0119/0.0.3f19/block/dasda
 	 * /sys/devices/css0/0.0.0000/0.0.0000/virtio0/block/vda
 	 */
-	if (sscanf(buf, "/sys/devices/css0/%*[0-9a-f.]/%[0-9a-f.]", busid) != 1)
+	if (sscanf(path, "/sys/devices/css0/%*[0-9a-f.]/%[0-9a-f.]", busid) != 1)
 		return -1;
 	return 0;
 }
--- a/ipl_tools/ccw.c
+++ b/ipl_tools/ccw.c
@@ -9,7 +9,6 @@
  * it under the terms of the MIT license. See LICENSE for details.
  */
 
-#include <assert.h>
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -17,6 +16,7 @@
 #include <unistd.h>
 
 #include "lib/util_path.h"
+#include "lib/util_panic.h"
 #include "ipl_tools.h"
 
 /*
@@ -26,9 +26,10 @@
  */
 static int device_sysfs_path(const char *device, char *path, const size_t path_size)
 {
-	assert(device);
-	assert(path);
-	assert(path_size == PATH_MAX);
+	util_assert(device != NULL, "Internal error: device is NULL");
+	util_assert(path != NULL, "Internal error: path is NULL");
+	util_assert(path_size == PATH_MAX, "Internal error: path_size is '%zu', but must be '%zu'",
+		    path_size, PATH_MAX);
 	char *buf = util_path_sysfs("block/%s/device", device);
 
 	if (!realpath(buf, path)) {
@@ -61,6 +62,26 @@ int ccw_is_device(const char *busid)
 	snprintf(path, sizeof(path), "/sys/bus/ccw/drivers/dasd-fba/%s", busid);
 	if (access(path, R_OK) == 0)
 		return 1;
+	return 0;
+}
+
+/*
+ * Check if the specified device is a valid virtio subchannel device
+ */
+int ccw_is_virtio_device(const char *device)
+{
+	char path[PATH_MAX] = { '\0' };
+	unsigned virtio = 0;
+
+	if (device_sysfs_path(device, path, sizeof(path)) != 0)
+		return -1;
+
+	/*
+	 * The output has the following format:
+	 * /sys/devices/css0/0.0.0000/0.0.0000/virtio0/block/vda
+	 */
+	if (sscanf(path, "/sys/devices/css0/%*[0-9a-f.]/%*[0-9a-f.]/virtio%u", &virtio) != 1)
+		return -1;
 	return 0;
 }
 
--- a/ipl_tools/cmd_chreipl.c
+++ b/ipl_tools/cmd_chreipl.c
@@ -391,7 +391,8 @@ static void dev_from_part_nvme(char *dev_name)
 static int set_reipl_type(const char *dev_name)
 {
 	if (strncmp(dev_name, "dasd", strlen("dasd")) == 0 ||
-	    strncmp(dev_name, "vd", strlen("vd")) == 0)
+	    strncmp(dev_name, "vd", strlen("vd")) == 0 ||
+	    ccw_is_virtio_device(dev_name) == 0)
 		l.reipl_type = REIPL_CCW;
 	else if (strncmp(dev_name, "sd", strlen("sd")) == 0)
 		l.reipl_type = REIPL_FCP;
--- a/ipl_tools/ipl_tools.h
+++ b/ipl_tools/ipl_tools.h
@@ -87,6 +87,7 @@ int nvme_is_device(char *fid_str, char *nsid_str);
  * CCW
  */
 extern int ccw_is_device(const char *devno);
+extern int ccw_is_virtio_device(const char *device);
 extern void ccw_busid_get(const char *device, char *devno);
 
 /*


nikita-dubrovskii:
This fixes an issue: https://github.com/coreos/fedora-coreos-tracker/issues/1434

borntraeger:
I think this makes sense.

mhartmay:
Does this work if `virtio-scsi-pci` is used for the SCSI controller?
E.g.
`-device virtio-scsi-pci,id=scsi0,bus=pci.0,addr=0x1`

I'm asking because I suspect it doesn't.  Here's the result of the realpath command executed in a guest using `virtio-scsi-pci`:
```
root@qemus390x:~# realpath /sys/block/sda/device
/sys/devices/pci0001:00/0001:00:00.0/virtio2/host0/target0:0:2/0:0:2:0
```

So the statement `strstr(resolved, "virtio") != NULL)` [1] would also evaluate to true, but the type is obviously not CCW.

If `virtio-scsi-ccw` is used the result of realpath is:
```
root@qemus390x:~# realpath /sys/block/sda/device
/sys/devices/css0/0.0.0000/0.0.0003/virtio0/host0/target0
```

[1] https://github.com/ibm-s390-linux/s390-tools/pull/154/files#diff-77c5828a786983ff0a5792636002c1aec766c63abb578610798c2c10bc4abc1bR400

nikita-dubrovskii:
@mhartmay thx for the observation! Indeed it won't work with `virtio-scsi-pci` :
```
$ realpath /sys/block/sda/device
/sys/devices/pci0001:00/0001:00:00.0/virtio0/host0/target0:0:0/0:0:0:0

$ chreipl /dev/sda
chreipl: Could not lookup device number for "sda"
```

mhartmay:
Code itself looks correct, but could you please refactor some parts of it? 
E.g. this part
```
snprintf(path, sizeof(path), "/sys/block/%s/device", dev_name);
if (realpath(path, resolved) == NULL)
		return -1;
```
is similar to
https://github.com/ibm-s390-linux/s390-tools/blob/master/ipl_tools/ccw.c#L83

In addition something similar to `sscanf` (e.g. https://github.com/ibm-s390-linux/s390-tools/blob/master/ipl_tools/ccw.c#L92) reads better than the `strstr()` construct.

nikita-dubrovskii:
@mhartmay Done

ipl_tools/cmd_chreipl.c
```diff
@@ -100,6 +100,7 @@ static const char *const usage_chreipl =
 "Options for node target:\n"
 "  Depending on underlying target type (ccw or fcp)\n";
 
+
```
mhartmay, line 103: This change seems to be unrelated.

ipl_tools/cmd_chreipl.c
```diff
@@ -390,8 +391,10 @@ static void dev_from_part_nvme(char *dev_name)
 
 static int set_reipl_type(const char *dev_name)
 {
+	char busid[10] = {'\0'};
 	if (strncmp(dev_name, "dasd", strlen("dasd")) == 0 ||
-	    strncmp(dev_name, "vd", strlen("vd")) == 0)
+	    strncmp(dev_name, "vd", strlen("vd")) == 0 ||
+	    ccw_busid_get_sysfs_new(dev_name, busid) == 0)
```
mhartmay, line 397: Unfortunately, I think this breaks list-directed IPL from ECKD DASD support. One fix would be to change the order... First check for `(strncmp(dev_name, "eckd", strlen("eckd")) == 0)
		l.reipl_type = REIPL_ECKD;` and then the checks dasd, vd etc...

@svens-s390

mhartmay: Secondly, do we need to consider `ccw_busid_get_sysfs_old` as well? Next possible issue, is the array length of bussid large enough? Do we need to protect against a buggy kernel?  I don't think it's a problem - but just mention it to get a second opinion.

nikita-dubrovskii: Now it becomes more complicated, because we  have to rely on `if ()` order, maybe it's better to use previous solution with `realpath ; strstr(virtio) && strstr(css)` ? I mean it's not a refactoring anymore, but some kind of breaking change

ipl_tools/ccw.c
```diff
@@ -17,6 +17,14 @@
...
 #include "ipl_tools.h"
 
+#define REALPATH(device, path)                                                \
+        do {                                                                  \
+                char buf[PATH_MAX] = {'\0'};                                  \
+                snprintf(buf, sizeof(buf), "/sys/block/%s/device", (device)); \
+                if (realpath(buf, (path)) == NULL)                            \
+                        return -1;                                            \
+        } while (0)
+
```
mhartmay, line 27: Hmm let's Sven decide if he likes this construct or not - I guess for two places I would leave it open-coded.

ipl_tools/ccw.c
```diff
@@ -42,6 +50,27 @@ int ccw_is_device(const char *busid)
...
+	char path[PATH_MAX] = {'\0'};
+	int virtio = -1;
+
+	REALPATH(device, path);
+
+	/*
+	 * The output has the following format:
+	 * /sys/devices/css0/0.0.0000/0.0.0000/virtio0/block/vda
+	 */
+	if (sscanf(path, "/sys/devices/css0/%*[0-9a-f.]/%*[0-9a-f.]/virtio%d", &virtio) != 1)
```
mhartmay, line 68:
```suggestion
	size_t virtio_num;

	REALPATH(device, path);

	/*
	 * The output has the following format:
	 * /sys/devices/css0/0.0.0000/0.0.0000/virtio0/block/vda
	 */
	if (sscanf(path, "/sys/devices/css0/%*[0-9a-f.]/%*[0-9a-f.]/virtio%zu", &virtio_num) != 1)
```

ipl_tools/ccw.c
mhartmay: ...and if `buf[PATH_MAX]` instead of `buf[4096]` is okay or not.

nikita-dubrovskii: Should be:
```
$ man -s3 realpath
...
If resolved_path is specified as NULL, then realpath() uses malloc(3) to allocate a buffer
of up to PATH_MAX bytes to hold the resolved pathname, and returns a pointer to this buffer.  
```

nikita-dubrovskii:
@mhartmay so, does PR now look ok?

ipl_tools/ccw.c
steffen-eiden: I'd prefer a more precise name for this macro: `REALPATH_BLOCK` for example.
Also, this could be a helper function instead of a macro.
`int realpath_block(const char* device, char* path)`

In any case please add a short description of what that function or macro does/

The compiler then will decide if it makes more sense to inline the function or not.

nikita-dubrovskii: Done, but i prefer macro in this case. Making a function would require more checks and probably memory allocation

ipl_tools/ccw.c
```diff
@@ -17,6 +17,17 @@
...
 #include "ipl_tools.h"
 
+/*
+ * Look up for the device in /sys/devices/ hierarchy.
+ */
+#define DEVICE_SYSFS_PATH(device, path)                                       \
+        do {                                                                  \
+                char buf[PATH_MAX] = {'\0'};                                  \
+                snprintf(buf, sizeof(buf), "/sys/block/%s/device", (device)); \
+                if (realpath(buf, (path)) == NULL)                            \
```
hoeppnerj, line 27: I wonder if it would make sense to put `realpath()` into `util_path_sysfs()` in general and then you wouldn't need this macro here. However, that needs some testing and maybe a bit more thoughts to not break other users.

ipl_tools/ccw.c
```diff
@@ -77,19 +109,16 @@ static int ccw_busid_get_sysfs_old(const char *device, char *busid)
...
-		return -1;
+	DEVICE_SYSFS_PATH(device, path);
 
 	/*
 	 * The output has the following format:
 	 * /sys/devices/css0/0.0.0119/0.0.3f19/block/dasda
 	 * /sys/devices/css0/0.0.0000/0.0.0000/virtio0/block/vda
 	 */
-	if (sscanf(buf, "/sys/devices/css0/%*[0-9a-f.]/%[0-9a-f.]", busid) != 1)
+	if (sscanf(path, "/sys/devices/css0/%*[0-9a-f.]/%[0-9a-f.]", busid) != 1)
```
hoeppnerj, line 121: Couldn't this function just call `util_sys_get_dev_addr()`?

https://github.com/ibm-s390-linux/s390-tools/blob/27708026d41006d8432a265d40c3c524c4047bf5/libutil/util_sys.c#L138

ipl_tools/ccw.c
```diff
@@ -17,6 +17,17 @@
...
+ * Look up for the device in /sys/devices/ hierarchy.
+ */
+#define DEVICE_SYSFS_PATH(device, path)                                       \
+        do {                                                                  \
+                char buf[PATH_MAX] = {'\0'};                                  \
+                snprintf(buf, sizeof(buf), "/sys/block/%s/device", (device)); \
+                if (realpath(buf, (path)) == NULL)                            \
+                        return -1;                                            \
+        } while (0)
+
```
hoeppnerj, line 30: Please use tabs for indentation.

ipl_tools/ccw.c
```diff
@@ -42,6 +53,27 @@ int ccw_is_device(const char *busid)
...
+	char path[PATH_MAX] = {'\0'};
+	size_t virtio = 0;
+
+	DEVICE_SYSFS_PATH(device, path);
+
+	/*
+	 * The output has the following format:
+	 * /sys/devices/css0/0.0.0000/0.0.0000/virtio0/block/vda
+	 */
+	if (sscanf(path, "/sys/devices/css0/%*[0-9a-f.]/%*[0-9a-f.]/virtio%zu", &virtio) != 1)
```
hoeppnerj, line 71: Considering my comment regarding testing, this makes it hard to test due to the hardcoded `/sys` path. However, just skipping `/sys` (using e.g. `strstr()`) wouldn't do the trick. We'd need a util sscanf() function that takes `SYSFS_ROOT` into account. Nothing you should do now. I just wanted to share my thoughts on it.

ipl_tools/ccw.c
```diff
@@ -42,6 +53,27 @@ int ccw_is_device(const char *busid)
...
 }
 
+/*
+ * Check if the specified device is a valid virtio device
+ *
+ * This does not work when booting from tape.
+ */
+int ccw_is_virtio_device(const char *device) {
+	char path[PATH_MAX] = {'\0'};
+	size_t virtio = 0;
```
hoeppnerj, line 63: Why are you using `size_t`? Shouldn't `unsinged int` be sufficient? `size_t` is typically used for count of bytes. It's a little confusing for this temporary variable here.

hoeppnerj (changes requested):
Hi @nikita-dubrovskii a couple more review comments from my side.

Also, please limit the commit message to lines to 72 characters. Our [editorconfig](https://github.com/ibm-s390-linux/s390-tools/blob/27708026d41006d8432a265d40c3c524c4047bf5/.editorconfig#L33) defines this limit too, please consider using it.

ipl_tools/ccw.c
nikita-dubrovskii: `size_t` is an `unsigned int`. I'm just used to it, cause lot of `stdlib` functions return `size_t/ssize_t`

ipl_tools/ccw.c
nikita-dubrovskii: Hmm, don't think so. `util_sys_get_dev_addr` would be overhead here, and existing code in this source file also uses `sscanf`. I don't want to make a refactoring here.

ipl_tools/ccw.c
nikita-dubrovskii: Done.  I guess changing of `util_path_sysfs` shouldn't be part of this PR, as its purpose is just to fix `virtio` block devices detection.

ipl_tools/ccw.c
```diff
@@ -15,8 +15,23 @@
...
+#include "lib/util_path.h"
 #include "ipl_tools.h"
 
+/*
+ * Look up for the device in /sys/devices/ hierarchy.
+ */
+#define DEVICE_SYSFS_PATH(device, path)                                                            \
+	do {                                                                                       \
+		char *buf = NULL;                                                                  \
+		buf = util_path_sysfs("block/%s/device", (device));                                \
```
mhartmay, line 27:
```suggestion
		char *buf = util_path_sysfs("block/%s/device", (device));                                \
```

ipl_tools/ccw.c
```diff
@@ -9,14 +9,33 @@
...
 #include <string.h>
 #include <unistd.h>
 
+#include "lib/util_path.h"
 #include "ipl_tools.h"
 
+/*
+ * Look up for the device in /sys/devices/ hierarchy.
+ */
+static char *device_sysfs_path(const char *device, char *path, size_t path_size)
```
mhartmay, line 25: Please don't return `char *`, because people might think they have to free the returned value. Instead, add to the documentation that path must be PATH_MAX large and the value will be replaced in place.

And please use `const size_t path_size`.

mhartmay:
LGTM. Acked-by: Marc Hartmayer <mhartmay@linux.ibm.com>

But someone more familiar with CSS/CCW should look at the functions names and documentation.

ipl_tools/ccw.c
```diff
@@ -9,14 +9,36 @@
...
 #include "ipl_tools.h"
 
+/*
+ * Look up for the device in /sys/devices/ hierarchy.
+ *
+ * path must be PATH_MAX large and the value will be replaced in place
+ */
+static int device_sysfs_path(const char *device, char *path, const size_t path_size)
+{
+	assert(device);
```
hbrueckner, line 29: I wonder whether `util_assert` should be used instead

nikita-dubrovskii: Done

dustymabe:
anything still blocking this?

hoeppnerj:
Pulled and integrated, will be pushed to master soon. Thanks!

PR #155, opened 2023-09-01
Fix missing semicolon

jcajka:
5.9.4 net-snmp started to require semicolon on the config_require there are no docs covering this change. Noticed via [RHBZ#2235734](https://bugzilla.redhat.com/show_bug.cgi?id=2235734)


SandyWinter:
Looks good to me, thank you for analyzing and providing this fix.

PR #158, opened 2023-10-12
consolidated s390 device configuration

steffen-maier:
Consolidate the persistent and dynamic configuration of s390-specific devices in Linux distributions by delegating the configuration to the existing framework `zdev` from s390-tools.

This pull request prepares s390 device configuration in initrds built by `dracut` for a Linux distribution independent mechanism.
Integrated in https://github.com/ibm-s390-linux/s390-tools/releases/tag/v2.31.0 (PR not merged but the individual commits were integrated).

Zdev's job is to perform low-level configuration after which the user gets architecture-independent objects such as block devices, SCSI devices, or network interfaces. Those can and should in turn be configured with existing common code mechanisms. So there's a clear separated layering for configuration duties.

In particular, the s390-specific devices currently are: DASD, ZFCP, and ZNET representing channel-attached network (QETH incl. OSA and HiperSockets, LCS, CTC). Zdev has a stable command line user interface and abstracts from sysfs and from a persistent configuration representation. Zdev encapsulates configuration details. Systems management code can simply delegate configuration to zdev and thus reduce architecture-specific code.

This improves user experience, serviceability, maintainability, and reduces test effort.

@ngueorguiev @sharkcz

### DIFF
--- a/zdev/dracut/95zdev-kdump/module-setup.sh
+++ b/zdev/dracut/95zdev-kdump/module-setup.sh
@@ -27,6 +27,7 @@ check() {
 
     # Ensure that required tools are available
     require_binaries chzdev lszdev || return 1
+    require_binaries sed || return 1
     require_binaries grep sort uniq || return 1
 
     return 0


steffen-maier:
Even though this is a draft pull request, I would appreciate review comments.
It's only a draft until we sorted out the dependencies and merge order of pull requests for different projects depending on this PR.

PR #163, opened 2024-03-13
correct mapping of ATTESTATION_NR when uvdevice doesn't support info IOCTL

VasiliyS:
UvDevice::ATTESTATION_NR was used to set the supported uv commands as a value directly, which led to *bit* 0 being set.

As a result, the library (UvDeviceInfo via Display)  reported that `Info` was supported.

### DIFF
--- a/rust/pv_core/src/uvdevice/info.rs
+++ b/rust/pv_core/src/uvdevice/info.rs
@@ -48,10 +48,14 @@ impl UvDeviceInfo {
         let mut cmd = uvio_uvdev_info::new_zeroed();
         match uv.send_cmd(&mut cmd) {
             Ok(_) => Ok(cmd.into()),
-            Err(crate::Error::Io(e)) if e.raw_os_error() == Some(libc::ENOTTY) => Ok(Self {
-                supp_uvio_cmds: (UvDevice::ATTESTATION_NR as u64).into(),
+            Err(crate::Error::Io(e)) if e.raw_os_error() == Some(libc::ENOTTY) => {
+            let mut supp_uvio_cmds : Lsb0Flags64 = 0u64.into();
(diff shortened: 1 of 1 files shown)


rust/pv_core/src/uvdevice/info.rs
```diff
@@ -48,10 +48,14 @@ impl UvDeviceInfo {
         let mut cmd = uvio_uvdev_info::new_zeroed();
         match uv.send_cmd(&mut cmd) {
             Ok(_) => Ok(cmd.into()),
-            Err(crate::Error::Io(e)) if e.raw_os_error() == Some(libc::ENOTTY) => Ok(Self {
-                supp_uvio_cmds: (UvDevice::ATTESTATION_NR as u64).into(),
+            Err(crate::Error::Io(e)) if e.raw_os_error() == Some(libc::ENOTTY) => {
+            let mut supp_uvio_cmds : Lsb0Flags64 = 0u64.into();
```
steffen-eiden, line 52: I'd prefer 
`let mut supp_uvio_cmds = Lsb0Flags64::default();`

steffen-eiden (changes requested):
Good catch, thanks for providing  fix.

Beside the little nit LGTM

steffen-eiden (changes requested):
Some additional nits regarding the commit message.

Could you add a short patch note what you are fixing.

also add a `Signed-off by: <Your Name> email` statement at the end of the commit message,

rust/pv_core/src/uvdevice/info.rs
VasiliyS: Ok, will do.

steffen-eiden:
Thank you for fixing the issue.

I will pick the patches, squash them and publish with the next wave of up-stream update.

steffen-eiden (approved):

PR #167, opened 2024-03-27
libpv: fix build with OpenSSL 1.1

sharkcz:
OpenSSL 1.1 seems to use a non-const parameter to `X509_name_dup()`, but `x509_armonk_locality_fixup()` is passing a const there. The compile then fails on "discards 'const' qualifier", when `-Werror` is used. Thus resolve with a type-cast like in `pv_c2b_name()`.


mhartmay:
Hi Dan, LGTM. While at it... can you please fix it for genprotimg as well? [1]

[1] https://github.com/ibm-s390-linux/s390-tools/blob/master/genprotimg/src/utils/crypto.c#L771

mhartmay:
Reviewed-by: Marc Hartmayer <mhartmay@linux.ibm.com>
for both patches.

PR #185, opened 2025-04-03
added VMLINK

trothr:
The VMLINK automounter script 
attempts to work similarly to the CMS 'VMLINK' command in that disks can be linked and mounted on-demand. 

See the CMS Help facility for information about the CMS 'VMLINK' command. 

The VMLINK automounter script works as follows: 

- define an automounter point under control of /etc/auto.vmlink such as /auto/vmlink 
- 'cd' to a sub-directory under that point such as /auto/vmlink/_myvm_._mydisk_ 

The disk in question will be linked (using 'vmcp link', read only) and mounted. 
By default, the first partition (or offset zero if the disk is not partitioned) will be mounted. 
Partitions can be indicated by a trailing ".1" or ".2" or ".3".

### DIFF
--- a/etc/auto.vmlink
+++ b/etc/auto.vmlink
@@ -1,0 +1,101 @@
+#!/bin/sh
+#
+#         Name: /etc/auto.vmlink
+#
+# AUTOVMLINK package, program script for the automounter
+# Rushal Verma <rusrushal13@gmail.com>
+# 2018-Aug-02 (Thursday, 19:42)
+#
+# Use this script as the program called by the automounter
+# for DASD which may be off-line when Linux is rebooted.
+# It will handle an automounter point (an automounter directory)
+# for which sub-directories represent mainframe disk I/O addresses.
+#
+
+# random check
+if [ "$*" = "*" ] ; then exit 1 ; fi
+
+DATENOW=`date "+# VMLINK %Y-%m-%d %H:%M:%S $*"`
+logger "$DATENOW"
+
+# arbitrary address
+SLOT=9000
+
+# conditionally source the config file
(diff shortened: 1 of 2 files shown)


hoeppnerj:
Hi @trothr this script seems like a nice helper script that might be useful for some people but is nothing we would provide through the s390-tools package, I'm sorry.
It also seems like it was authored by someone else in 2018? What is this about?
```
# Rushal Verma <rusrushal13@gmail.com>
# 2018-Aug-02 (Thursday, 19:42)
```

trothr:
I should have been more clear. 
The original "automounter for z/VM minidisks" was done when I worked for a large enterprise company. 
Rushal Verma served as a intern during the 2018 season and re-worked the basic script, making it more general, 
under my direction. We released it as its own project. 

Recently in a "VM Council" call, I asked about general adoption of this VMLINK automounter. 
It was recommended that I/we contribute it to S390 Tools. I conferred with Rushal and he agreed. 

The stand-alone works fine and is available as an RPM. 
But it seems silly to have an RPM for just the one script (and optional config file). 

On-demand sharing of disks a a normal thing for guest operating systems on z/VM. 
But with Linux it is less well known and poorly understood. I would hope that this VMLINK automounter 
would make sharing disks among Linux guests both normal and seamless. 

Thanks!

trothr:
so ... can you please add this script to S390 Tools?

hoeppnerj:
Hi @trothr , as mentioned already above, this is considered a useful helper script that uses s390-tools provided tools. It represents a certain workflow and might be helpful for some people. There are probably tons of scripts of this nature that represent specific workflows and are helpful to certain people. Please understand that we can't integrate scripts like this into the package.

PR #186, opened 2025-05-12
zipl: remove global sync() call on exit

nikita-dubrovskii:
zipl already uses `fsync()` and `ioctl(BLKFLSBUF)` to flush all required disks. The extra `sync()` is redundant and can block if NFS mounts are unreachable.

Issue: https://github.com/openshift/os/issues/1720


edward-ibm:
Nop-nop.

The fsync() calls you mentioned are invoked only at installation time, accept only file descriptors of base disks and are to write boot records only.

However, besides boot records, zipl jobs create numerous boot data (like bootmap files). It relies exactly on that global sync(), which is obviously *not* redundant.

We can get rid of that global sync(), replacing it with fsync() calls only against the files that zipl(8) writes to, but it would be not a 2-lines patch.

Thanks,
Eduard.

zipl/src/install.c
```diff
@@ -291,7 +291,7 @@ static int install_eckd_cdl_ld(int fd, disk_blockptr_t *br,
 		return -1;
 	rc = DRY_RUN_FUNC(misc_write(fd, &vl, sizeof(vl)));
 	if (rc)
-		error_text("Could not update volume lablel 'vol1'");
+		error_text("Could not update volume label 'vol1'");
```
hoeppnerj, line 294: Please don't intermix typo changes with functional improvements or bug fixes. The context is different. Please provide a separate patch for that. Thanks!

nikita-dubrovskii: fixed

edward-ibm (commented):
This is not enough.
Please, carefully inspect all cases of open(). Don't forget about all zipl helpers (the helper for for device-mapper, as well as the helpers for ngdump provided for initramfs-tools and for dracut).

Thanks,
Eduard.

nikita-dubrovskii:
> This is not enough. Please, carefully inspect all cases of open(). Don't forget about all zipl helpers (the helper for for device-mapper, as well as the helpers for ngdump provided for initramfs-tools and for dracut).

I guess now all `open(O_RDWR)` cases within `zipl` are covered. Regarding other helpers and tools - i'll check extra.

edward-ibm:
I have sent the latest version of the fixup (3 pathes) by email.
Everyone is cc-ed.

Thanks,
Eduard.

PR #190, opened 2025-09-09
Fix zipl EBCDIC code page 500 conversion and decrease size by 200 bytes

huth:
The tables ebcdic_037 and ebcdic_500 are currently exactly the same, though the 500 should differ at 7 entries (see the tables in the Linux kernel or https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=ebcdic-code-point-differences-between-ccsids for example). Fix it by removing the ebcdic_500 table and by patching the other table during start() if necessary. That way we can safe also about 200 bytes in the stage3.bin binary.

Please review & test carefully since I only have limited access to machines for testing here (e.g. no z/VM available on my side).


steffen-eiden (changes requested):
Thanks for the patch Thomas.
With my nits fixed:

`Acked-by: Steffen Eiden <seiden@linux.ibm.com>`

PR #191, opened 2025-10-16
 lib(ekmfweb|kmipclient): Use ln without -r

sertonix:
The option isn't portable (not in POSIX or busybox) and it easily works without

### DIFF
--- a/libekmfweb/Makefile
+++ b/libekmfweb/Makefile
@@ -91,13 +91,13 @@ libekmfweb.so.$(VERSION): ALL_LDFLAGS += -shared -Wl,--version-script=libekmfweb
 	-Wl,-z,defs,-Bsymbolic -Wl,-soname,libekmfweb.so.$(VERM)
 libekmfweb.so.$(VERSION): ekmfweb.o utilities.o cca.o $(libs)
 	$(LINK) $(ALL_LDFLAGS) $^ $(LDLIBS) -o $@
-	ln -srf libekmfweb.so.$(VERSION) libekmfweb.so.$(VERM)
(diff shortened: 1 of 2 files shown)


ifranzki:
I agree that for `make install` as such it should work well without `-r`, i.e. using absolute symlinks.

However, I wonder what happens when this is used during packaging by the distros?
I.e. for RPM build, `make install` is run with a target directory somewhere inside `rpmbuild/BUILDROOT/<package>/`, so the symlinks would point to `rpmbuild/BUILDROOT/<package>/....../libekmfweb.so`, i.e. the absolute path of the library.
What happens when those symlinks are then packaged into the RPM? Do they keep the absolute path? If so, then they won't work later when the RPM is installed on a different system, where no rpmbuild directory structure exists.

sertonix:
>  so the symlinks would point to rpmbuild/BUILDROOT/<package>/....../libekmfweb.so, i.e. the absolute path of the library.

The symlinks are still relative. Without the `-r` option you specify the symlink target string directly instead of relying on some logic in `ln` to construct a relative path. In your example the target will still be `libekmfweb.so` just like before

ifranzki:
> > so the symlinks would point to rpmbuild/BUILDROOT//....../libekmfweb.so, i.e. the absolute path of the library.
> 
> The symlinks are still relative. Without the `-r` option you specify the symlink target string directly instead of relying on some logic in `ln` to construct a relative path. In your example the target will still be `libekmfweb.so` just like before

OK, that makes sense. Then all is good. I'll approve and @hoeppnerj can pull the changes.

ifranzki (approved):

PR #192, opened 2025-10-16
tools: Fix compilation with musl and gcc 14

amyspark:
Hi all,

This MR fixes deploying s390-tools on Musl-based systems, like Alpine. These need an extra include:

```
lscss.c: In function 'is_sch_vfio':
lscss.c:392:20: error: implicit declaration of function 'basename' [-Wimplicit-function-declaration]
  392 |         if (strcmp(basename(driver_path), "vfio_ccw") == 0)

device.c: In function 'device_read_active_attrib':
device.c:426:45: error: implicit declaration of function 'basename'; did you mean 'rename'? [-Wimplicit-function-declaration]
  426 |                         value = misc_strdup(basename(link));
      |                                             ^~~~~~~~
      |                                             rename
```

I'd sent this patch to Alpine ages ago ([see this MR](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/70132)) but I forgot to upstream it.

Let me know what you think.

PR #193, opened 2025-10-28
Fix compilation on Alpine linux (and some more minor fixes)

sertonix:
This patchset has become quiet large. Would it be better to split this into separate PRs?

Aside from this, https://github.com/ibm-s390-linux/s390-tools/pull/191 and https://github.com/ibm-s390-linux/s390-tools/pull/192 there are 2 build failures left but I don't think they can be fixed well here:
- `argz.h` doesn't exist on musl libc. Alpine is and probably will continue to use a copy of `argz.h` and `argz.c` from glibc
- `feenableexcept` doesn't exist on musl libc. Alpine has currently disabled `cpuplugd` cause of that. It might be possible to do something similar to https://github.com/ArduPilot/ardupilot/pull/11871

### DIFF
--- a/opticsmon/ethtool.c
+++ b/opticsmon/ethtool.c
@@ -1,4 +1,3 @@
-#include <execinfo.h>
 #include <sys/wait.h>
 #include <sys/socket.h>
 #include <fcntl.h>


steffen-eiden:
the `feenableexcept` approach seems reasonable IMO.

steffen-eiden:
> Thanks for the review!
> 
> > `Reviewed-by: Steffen Eiden <seiden@linux.ibm.com>`
> 
> Should I amend the commit message with that?

yes, thank you!

steffen-eiden (approved):

include/boot/s390.h
hoeppnerj: Regarding the commit message for this change:
- Please use imperative mood (no "we", "us", "our", etc)
- You can remove the comment regarding the renaming as an alternative
- Please reference commits in this form: `11bdab26297e ("include/boot/s390.h: add guard for struct __vector128")` (also use a commit abbrev value of 12)

dasdinfo/dasdinfo.c
```diff
@@ -388,7 +388,7 @@ dinfo_is_busiddir(const char *fpath, const struct stat *UNUSED(sb),
 
 	if (tflag != FTW_D || (strncmp((fpath + ftwbuf->base), searchbusid,
 				       strlen(searchbusid)) != 0))
-		return FTW_CONTINUE;
+		return 0;
```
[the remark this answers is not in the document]
sertonix: Done

include/boot/s390.h
sertonix: Done

PR #195, opened 2025-11-19
fdasd: Add non-interactive script friendly interface

vojtechtrefny:
This is just a work in progress patch, mostly giving you a chance to tell me you are not interested in this and/or suggest a better approach.
The idea is to provide a script friendly interface similar to `parted --script`.  Calling this can look for example like

```
# fdasd --script "d 1 a 2 +100m w" /dev/dasdc
```
to remove an existing partition and add a new 100 MiB one. 

I am trying to reuse as much of the existing code as possible so this hopefully shouldn't create more maintenance work in the future.
This is obviously missing other commands, documentation and needs some more testing etc, as I said, still work in progress and mostly looking for some early feedback now.

### DIFF
--- a/fdasd/fdasd.8
+++ b/fdasd/fdasd.8
@@ -15,6 +15,10 @@ command line mode:
      \fBfdasd\fR [\-s] [\-r] [\-C] {\-a[\-k|\-l \fIvolser\fR]|\-i|\-p|\-c \fIconf_file\fR}
 [-f \fI[type,blocksize]\fR] \fIdevice\fR
 .br
+script mode:
+.br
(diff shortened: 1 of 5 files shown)


steffen-eiden:
Hey, this sounds good.I had a quick glance looks good.

I know this is just an early draft, but for the real deal please split up the patch into multiple and first do all the refactoring (e.g. move stuff to separate functions) ans then the actual feature addition.|

I am not soo fluent with the dasd tools but you may be able to reuse the CLI parser to parse you script.
(under the assumption that the script mode cannot do more than the interactive)

@hoeppnerj  may provide more information and/or feedback

vojtechtrefny:
Rebased to the latest master.

PR #199, opened 2026-02-05
scripts/cpictl: fix RedHat CoreOS detection

nikita-dubrovskii:
CoreOS uses VARIANT_ID instead of a unique ID in `/etc/os-release`.
Extend distro detection to identify RHCOS, other CoreOS variants
may be added later if needed.

RHCOS is showing in HMC as RHEL, because `system_level` is:
```
[core@cosa-devsh ~]$ cat /sys/firmware/cpi/system_level
0x010906023a050e00
```
But should be:
```
[core@cosa-devsh ~]$ cat /sys/firmware/cpi/system_level
0x070906023a050e00
```

This PR also fixes `shellcheck` warings.

Issue: https://jsw.ibm.com/browse/OCPVIP-1471

### DIFF
--- a/scripts/cpictl
+++ b/scripts/cpictl
@@ -137,7 +137,7 @@ fail_with()
 {
 	echo "$1" >&2
 	echo "Try '$PRG --help' for more information." >&2
-	exit ${2:-$EXIT_FAILURE}
+	exit "${2:-$EXIT_FAILURE}"
 }
 
 cpi_commit()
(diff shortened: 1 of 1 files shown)


oberpar (commented):
Code looks good. Comments seem rather verbose - not sure if the github links are necessary, or the mentioning of distributions that are not supported. Beyond that I have no objections.

PR #204, opened 2026-06-08
Use EVP_PKEY_id() for OpenSSL 1.1 detection

sgallagher:
The previous function, EVP_PKEY_meth_remove(), was deprecated in OpenSSL 3.0 and fully removed in OpenSSL 4.0. As a result, it fails detection of OpenSSL 1.1+ if OpenSSL 4 is present.


ifranzki (approved):
Good catch.

LGTM.

steffen-eiden (approved):
`Reviewed-by: Steffen Eiden <seiden@linux.ibm.com>`

PR #209, opened 2026-08-04
pvebc: Fix executable permission on sel-ebc-modules.conf

frank-heimes:
`sel-ebc-modules.conf` is a plain text kernel modules list and should not be installed as executable.

In `rust/Makefile`, it was previously bundled in the `-m 755` `install` command alongside the shell scripts. Move it to a separate `-m 644` `install` call.


fcallies:
Good catch! I would prefer adding it into the existing `-m 644` block which is one line below your insertion ;)
There is no need for its own install block.

rust/Makefile
```diff
@@ -115,7 +115,8 @@ ifneq ($(HAVE_DRACUT),0)
 		pvebc/$(SEL_EBC_MODDIR)/override-crypttab.sh \
 		pvebc/$(SEL_EBC_MODDIR)/pvebc-wrapper.sh \
 		pvebc/$(SEL_EBC_MODDIR)/boot-mount.sh \
-		pvebc/$(SEL_EBC_MODDIR)/sel-ebc-modules.conf \
+		$(DESTDIR)$(DRACUTMODDIR)/$(SEL_EBC_MODDIR)
+	$(INSTALL) -m 644 pvebc/$(SEL_EBC_MODDIR)/sel-ebc-modules.conf \
 		$(DESTDIR)$(DRACUTMODDIR)/$(SEL_EBC_MODDIR)
 	$(INSTALL) -m 644 pvebc/$(SEL_EBC_MODDIR)/sel-ebc-boot-mount.service \
```
fcallies, line 119: You can add it exactly here instead. This `-m 644` install block already contains ebc resources so it fits in perfectly

frank-heimes:
Sure - updated the PR: sel-ebc-modules.conf is now part of the existing -m 644 block, the extra install block is gone.

fcallies (approved):
LGTM, thank you!

PR #210, opened 2026-08-04
cpumf/lspai: Fix whatis entry in lspai.8 man page

frank-heimes:
`lexgrog` cannot expand the custom string macro `\*c` in the `.SH NAME` section and therefore fails to parse the whatis entry for `lspai(8)`.

Replace the macro `\*c` with the literal program name `lspai` in the `.SH NAME` line only.


steffen-eiden:
Could you replace all occurrences of that macro and remove it while you are at it?

frank-heimes:
Sure - updated the PR: all occurrences of the macro are now replaced and its definition is removed.

hoeppnerj:
Looks good, thanks!

